PR #13, opened 2017-04-19
[fix] icon typo

jeremie-lucca:


lucienbertin:
you fixed `hearth` -> `heart` in the demo but did you changed the mapping of the icon in the font/scss ?

jeremie-lucca:
The font ligature was setted on `heart`, so only the demo was wrong. 
It looks that if icon's text start by the ligature string, the icon will be well rendered. That why `hearth` worked on the demo.

PR #29, opened 2017-05-04
[enh] cut demo

jeremie-lucca:
It's a first step.
I needed a simple way to cut the demo in sections so I tried vue.js.

... and templates are back.


jeremie-lucca:
That's right. 
It's only a first step for the moment. 

I have to search what could be efficient. 
Maybe a wiki ?

PR #30, opened 2017-05-04
add link script

lucienbertin:
[npm link](https://docs.npmjs.com/cli/link) is the way to wwork on several interdependnat npm packages.

here we will have several packages with dependencies between them, `ng` is dependant on `scss`. but in the future we'll have
`scss depends on `icons`
`webcomponents` depends on `icons` and `scss`
`ng` depends on `icons`, `scss`, `webcomponents`

this script does all the necessary linking

### DIFF
--- a/tools/link.js
+++ b/tools/link.js
@@ -1,24 +1,17 @@
 const npmRun = require('npm-run');
 
+// respect dependency order here
 const packages = [
-	// { name: 'scss', dependencies: ['icons'] },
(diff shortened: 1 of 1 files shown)


lucienbertin:
@GuillaumeNury do you think i should call it in the `post install` of the root dir ?

GuillaumeNury:
Yes it should be the only thing to run in the root level

PR #32, opened 2017-05-05
Ux137 ng states

jeremie-lucca:
With demo & styleguide refactoring, this branch is generating a lot of conflicts. 
I think it would be a good idea to merge this part now. 
(It's about create mixin which could be use by ng classes)


packages/scss/src/mixins/_loading.scss
```diff
@@ -0,0 +1,45 @@
...
+
+	&:before{
+		border: 2px solid rgba(0,0,0,.2);
+		border-radius: 50%;
+		bottom: 0;
+		content: '';
+		display: block;
+		height: $size;
+		left: 0;
+		line-height: .8rem;
```
sanlucca: Doesn't line-height depend on $size too ?

jeremie-lucca: I tried with 5 different button sizes and it was ok... :)

lucienbertin:
@sanlucca @jeremie-lucca xan i merge this ?

jeremie-lucca:
@lucienbertin Yep!

PR #33, opened 2017-05-09
rem link script, do all the linking in postinstall script at root

lucienbertin:


GuillaumeNury:
Why deleting this scripts ? It should be ok for juste one dependancy, but it will become a pain in the future

lucienbertin:
the problem is that you cant call `npm link` before doing a `npm install` it will crash

but i can rework the link script and rename it so it calls 

```
npm link dependencies
npm install
npm link me
```
package after package with respect to each dependency

lucienbertin:
@GuillaumeNury i did wot i said i would do

Issue #38, opened 2017-05-12
scss - convert $variables to the theming


lucienbertin:
@jeremie-lucca for you

jeremie-lucca:
Done here : https://github.com/LuccaSA/lucca-front/pull/39

lucienbertin:
@jeremie-lucca in the PR, you can add `fixes #<issue number>` and the issue will be closed when the pr merges

PR #39, opened 2017-05-15
Scss.ux208.variables in theming

jeremie-lucca:


packages/scss/src/theming/_colors.scss
```diff
@@ -1,8 +1,11 @@
...
-$luTheme: _set($luTheme, 'palette.colors', $colors);
-
-
+	primary: #FFCC00,
+	secondary: #FF6600,
+	success: #5CB85C,
+	warning: #FF9900,
+	error: #FF3300,
+	greyStandard: #666666,
+	greyLight: #999999,
```
lucienbertin: you could have `grey` and `light` instead of `greyStandard` and `greyLight`

packages/scss/src/_typo.scss
```diff
@@ -7,39 +7,39 @@ html {
...
 	vertical-align: middle;
 }
 
 h1 {
-	font-size: $font-biggest;
+	font-size: _get($luTheme, "font.size.biggest");
 	font-weight: 600;
-	line-height: $font-line-biggest;
+	line-height: _get($luTheme, "font.line.biggest");
 	margin: 0 0 1.7rem;
```
bpaugam: Rem units could be dangerous, as you may not control body font-size (say the lucca banner suddenly decides to change it, or any other third-party component stylesheet). That was why a `lui_rem(n)` function was used in lucca-ui: https://github.com/LuccaSA/lucca-ui/blob/97640717facdcca8f51361d0068e581a0e8c9e8b/scss/core/utilities/theming/_theming.scss#L32

packages/scss/src/_typo.scss
```diff
@@ -7,39 +7,39 @@ html {
 }
 
 body {
-	background: $greyBackground;
-	color: $grey;
+	background: #F5F5F5;
+	color: _get($luTheme, "palette.colors.greyStandard");
```
bpaugam: As you repeat `_get($luTheme, "path")` everywhere, a `_theme("path")` shorthand function could be useful.

lucienbertin:
i added a function `_theme($keys)` that is equivalent to `_get($luTheme, $keys)`

packages/scss/src/components/_button.scss
```diff
@@ -52,97 +52,89 @@
...
 		&:active {
-			background: darken($greyLighter, 5%);
+			background: darken(_get($luTheme, "palette.colors.greyLight"), 5%);
 		}
 	}
 
 	&.mod-error {
-		background: $red;
+		background: _get($luTheme, "palette.colors.error");
 		color: white;
```
bpaugam: How can you be so sure the error color I'll set in my overrides isn't too pale for white text to be illegible over it?
In Lucca-ui, a palette color was a main color as well as the color text should have when writter over such a color.

packages/scss/src/components/_card.scss
```diff
@@ -27,6 +27,6 @@ $cardPadding: 1.5rem;
 
 .card {
 	&.mod-grey {
-		background: $greyBackground;
+		background: #F5F5F5;
```
bpaugam: Should this really be hard-coded ?

packages/scss/src/components/_checkbox.scss
```diff
@@ -9,8 +9,8 @@
 
 	&:checked ~ .checkbox-label {
 		&::before {
-			background-color: $primary;
-			box-shadow: inset 0 0 0 2px $primary;
+			background-color: _get($luTheme, "palette.colors.primary");
+			box-shadow: inset 0 0 0 2px _get($luTheme, "palette.colors.primary");
```
bpaugam: Why is "colors" plural, but not "palette" ?

packages/scss/src/components/_container.scss
```diff
@@ -1,21 +1,21 @@
 .container {
 	margin: 0 auto;
 	padding: 0 2rem;
-	width: $xl - 40px;
+	width: _get($luTheme, "breakpoints.xl") - 40px;
 
-	@media (max-width: $xl) {
-		width: $lg - 40px;
+	@media (max-width: _get($luTheme, "breakpoints.xl")) {
```
bpaugam: Lucca-ui mixins `lui_screens_larger_than()` and `lui_screens_smaller_than()` were kinda useful for developers always forgeting media queries syntax. Only problem was, it wasn't clear wether it was strict or not...

packages/scss/src/components/_container.scss
```diff
@@ -1,21 +1,21 @@
 .container {
 	margin: 0 auto;
 	padding: 0 2rem;
-	width: $xl - 40px;
+	width: _get($luTheme, "breakpoints.xl") - 40px;
```
bpaugam: Where is this `40px` coming from?

packages/scss/src/components/_container.scss
```diff
@@ -26,34 +26,34 @@
...
 	}
 
 	&.mod-xl {
-		max-width: $xl;
+		max-width: _get($luTheme, "breakpoints.xl");
 	}
 
 	&.mod-navSide {
-		width: $xl - 170px;
+		width: _get($luTheme, "breakpoints.xl") - 170px;
```
bpaugam: And now `170px`?

packages/scss/src/components/_grid.scss
```diff
@@ -131,7 +131,7 @@
 
 /* SM - Columns */
 
-@media only screen and (min-width: $sm) {
+@media only screen and (min-width: _get($luTheme, "breakpoints.sm")) {
 	.grid-sm {
 		flex-basis: 0;
```
bpaugam: I've read using the `flex: none | [ <'flex-grow'> <'flex-shrink'>? || <'flex-basis'> ]` shorthand rather than `flex-grow`, `flex-shrink`... properties is recommended.

bpaugam: Plus we ain't supporting old versions of browsers, is that assumed?

packages/scss/src/components/_input.scss
```diff
@@ -19,21 +19,21 @@
 	width: 100%;
 
 	&::placeholder {
-		color: $greyLighter;
+		color: #CCCCCC;
```
bpaugam: Should this be hard-coded?

packages/scss/src/components/_label.scss
```diff
@@ -1,9 +1,9 @@
...
 	background: #CCCCCC;
 	border-radius: 3px;
-	color: $grey;
+	color: _get($luTheme, "palette.colors.greyStandard");
 	display: inline-block;
-	font-size: $font-small;
+	font-size: _get($luTheme, "font.size.small");
 	font-weight: 400;
 	line-height: 1;
 	margin: 0 .3rem;
```
bpaugam: If every measure is rem based, that means I won't be able to make a "fat" input in my project.

packages/scss/src/components/_switch.scss
```diff
@@ -10,7 +10,7 @@
 }
 
 .switch-label {
-	color: lighten($grey, 25%);
+	color: lighten(_get($luTheme, "palette.colors.greyStandard"), 25%);
```
bpaugam: Shouldn't this be your `greyLight` color? Or maybe a specific variable?

packages/scss/src/components/_table.scss
```diff
@@ -1,8 +1,8 @@
 .table {
-	border-bottom: 1px solid $greySeparator;
+	border-bottom: 1px solid _get($luTheme, "commons.divider");
```
bpaugam: The `divider` variable being a color is not intuitive

packages/scss/src/components/_textarea.scss
```diff
@@ -59,15 +59,15 @@
...
 		box-shadow: none;
-		color: $greyLighter;
+		color: _get($luTheme, "palette.colors.greyLight");
 		cursor: not-allowed;
 	}
 }
 
 .textarea {
 	&.is-filled {
 		.textarea-label {
```
bpaugam: Why do you have to have distinct classes for an input and a textarea label? Is this the case for selects as well?

packages/scss/src/components/_tooltip.scss
```diff
@@ -1,14 +1,14 @@
...
 	color: white;
 	display: inline-block;
-	font-size: $font-small;
+	font-size: _get($luTheme, "font.size.small");
 	padding: .5rem 1rem;
 	position: relative;
 
 	&::before {
-		border-color: $grey transparent transparent transparent;
+		border-color: _get($luTheme, "palette.colors.greyStandard") transparent transparent transparent;
```
bpaugam: The triangle making mixin in Lucca-ui could have been kinda useful here. Same reason, many developers won't remember the css triangle trick.

packages/scss/src/functions/_get.scss
```diff
@@ -0,0 +1,25 @@
+@function _get($keys) {
+	$keys: str-explode($keys, '.');
+
+	$result: ();
+	$themeVar: nth($keys, 1);
+	@if ($themeVar == 'lui') {
```
bpaugam: ???

packages/scss/src/mixins/_loading.scss
```diff
@@ -1,6 +1,6 @@
 @mixin loading($size) {
-	&:before {
-		border: 2px solid rgba(0,0,0,.2);
+	&::before {
+		border: 2px solid rgba(0, 0, 0, .2);
```
bpaugam: What about a svg loader? That sure would be cool.

packages/scss/src/theming/_breakpoints.scss
```diff
@@ -1,10 +1,11 @@
...
-	xs:  576px,
-	sm:  768px,
-	md:  992px,
+	xs: 576px,
+	sm: 768px,
+	md: 992px,
 	lg: 1200px,
 	xl: 1350px,
 	xxl: 1500px,
 	xxxl: 1650px,
```
bpaugam: Are these arbitrary? If they are device sizes, shouldn't they be called as such ?

packages/scss/src/theming/_spacing.scss
```diff
@@ -0,0 +1,10 @@
+$spacing: (
+	tiny: 7px,
+	small: 15px,
+	standard: 30px,
+	big: 45px,
+	bigger: 60px,
+	biggest: 90px,
```
bpaugam: What are these ?

packages/scss/src/utilities/_spacing.scss
```diff
@@ -10,105 +10,105 @@
 /* Margin bottom */
```
bpaugam: The horror! One-property-classes! Proudly bearing the hated `!important` prefix on top of that!
Is this really needed ?

packages/scss/src/_typo.scss
jeremie-lucca: I'm not sure to understand the problem. 
Lucca Front set the body to 16px so all components are relative to this value.
If someone decide to change the body's font value, each components font will also change.
So, lucca banner fonts are also relative to the body font. If we want to reduce or increase the banner font size we will be able to set the value inside a "banner" scss/src/components file.

packages/scss/src/components/_button.scss
jeremie-lucca: Seems a good idea. I write it down in my backlog.

packages/scss/src/components/_card.scss
jeremie-lucca: For the moment I'm not sure if this color (which is used as body background color & in card component) should be defined on theming > commons or separately. 
Any advice?

packages/scss/src/components/_container.scss
jeremie-lucca: I set a -40px (which is a totaly arbitrary value) to have a container alway less wide than the view port, under each breakpoint. 

So it's only a way to define the container width relatively from our breakpoints.

packages/scss/src/components/_container.scss
jeremie-lucca: I'd rather let the CSS simple. But if it's really needed, I'm ok with it.

packages/scss/src/components/_container.scss
jeremie-lucca: As it's a value relative to the nav-side component (the aside main menu), it should be in a variable. 
Do you think that the nav-side component width should be also overridable?

packages/scss/src/components/_grid.scss
jeremie-lucca: As I saw many flex usages on some BU I thought it was ok. 
So... what's our IE compatibility requirements?

packages/scss/src/components/_input.scss
jeremie-lucca: I'd say yes... It's a very specific color usage. And I'm not sure that it's a got idea to let the placeholder color be overrided.

packages/scss/src/components/_label.scss
jeremie-lucca: Some labels mods will be added in the future (as it's the case for buttons). 
I prefere elements with quantified sizes to have a consistent UI and reduce the CSS for each BU.

packages/scss/src/components/_table.scss
jeremie-lucca: I agree. What should you suggest?

packages/scss/src/components/_textarea.scss
jeremie-lucca: Yep! 

With the methodology I follow each Sass components have to be separated. 
If some code is used in my places, we can use a mixin. 
But as it lead to less clarity I think it's okay to have a bit of duplicated code in 2 or 3 files in this case.

packages/scss/src/components/_table.scss
bpaugam: Probably something like this?
```scss
$luTheme: (
  commons: (
    divider: (
      weight: 1px,
      style: solid,
      color: _theme("palette.colors.grey.color")
    )
  )
);

@function _getDividerOneLiner() {
  @return "#{_theme('commons.divider.weight')} #{_theme('commons.divider.style)'} #{_theme('commons.divider.color)'}";
}

.table {
  border-bottom: _getDividerOneLiner();
}
```

packages/scss/src/theming/_breakpoints.scss
jeremie-lucca: As some tablets are bigger than laptops I prefere a "size relative" naming. 
Furthermore, how should we call biggers breakpoints?

packages/scss/src/theming/_spacing.scss
jeremie-lucca: Predefined values that can be used for margins and paddings.

packages/scss/src/_typo.scss
bpaugam: I personally never trust anything or anyone. I always live in fear of having something changing body font-size, thus destroying my measures.

packages/scss/src/components/_container.scss
bpaugam: You could add a `spacing` property to your breakpoints map?
```scss
$breakpoints: (
  xl: (
    breakAt: 1600px,
    spacing: 40px
  )
);

.container {
  width: _get($luTheme, "breakpoints.xl.breakAt") - _get($luTheme, "breakpoints.xl.breakAt");
}
```

packages/scss/src/components/_container.scss
bpaugam: Sure thing! Some like it big, others small!

packages/scss/src/components/_grid.scss
bpaugam: For flex, I can't remember where I saw that, but that was prolly browser-support related.

Regarding browser support, I guess that depends on the application. We (Timmi) should support IE 10 and up.

packages/scss/src/components/_input.scss
bpaugam: What about making it rgb, in case it's placed over a dark background?

packages/scss/src/utilities/_spacing.scss
jeremie-lucca: I use !important ONLY for utilities. 
As utilities are only called as "u-" classes on the DOM, there goal are to override components value for basics infos for UI requirements like margins, padding, text color, alignment, etc. 
It's the way bootstrap use "pull-left" or "pull-right" for exemple.

It's an useful way to avoid writing many CSS lines for basic UI usages.

sanlucca: But are they useful when you have people dedicated to styling ? Isn't is risky to offer such tools that can turn into "inline styling" ?

packages/scss/src/_typo.scss
sanlucca: I think we have to be careful of where we use em & rem. I understand why Benoit wants em as it allows to enlarge or reduce a whole component based on one font-size. We might have to be specific on where em are useful and where it should be avoided.

packages/scss/src/components/_switch.scss
jeremie-lucca: Yup, fixed.

packages/scss/src/utilities/_spacing.scss
jeremie-lucca: I understand, but I used these with my previous team during 1 or 2 years, and it was a huge time saving ad it avoid creating CSS sheets for many pages.

By removing them, it will lead to add more classes or modifiers, which mean that your components will have style in multiple style sources, especially if you have differents margins to apply on the same component for UI requirements. 

I'm pretty sure that we will lose time without them, and as it's only helpers, it's absolutely not an obligation to use it.

packages/scss/src/components/_container.scss
jeremie-lucca: Fixed

packages/scss/src/components/_container.scss
jeremie-lucca: Fixed

packages/scss/src/components/_grid.scss
jeremie-lucca: I could create a grid modifier de switch between a standard grid and flex.
Easy to do but quite big to write. --> in my backlog

packages/scss/src/components/_tooltip.scss
jeremie-lucca: Yep, it would be an useful mixin. I'll try to create it the next time it will be needed.

packages/scss/src/components/_table.scss
jeremie-lucca: Fixed

jeremie-lucca:
I made some improvements. 
I keep some in my backlog, some should lead to more discussions so I think we can merge at this point.

I hope it's ok for you all!

lucienbertin:
fixes #38

PR #41, opened 2017-05-19
[add][ux222] disabled buttons

jeremie-lucca:


packages/scss/src/components/_button.scss
```diff
@@ -154,6 +154,17 @@
 // ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
 
 .button {
+
+	/* Disabled */
+
+	&[disabled] {
```
sanlucca: Wouldn't it be useful to also add a .is-disabled class just in case ?

packages/scss/src/components/_button.scss
```diff
@@ -153,11 +155,28 @@
 // STATES
 // ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
 
+$buttonDisabled: .5;
```
lucienbertin: i think this sort of variables should be in the theme

jeremie-lucca: Right. As I think this opacity could be used in multiple components, I set it in "commons.disabled.opacity".

PR #46, opened 2017-05-23
[ref] palette color/text

jeremie-lucca:
I set a combo of background/text for each palette's colors, but I have a small issue with text colors. 

The grey and light colors are currently used for body texts AND in some components backgrounds like buttons, labels, or even switch toggle. This is a problem, because we can't dissociate colors for these texts and backgrounds. 

So, should text colors (standard grey, and light grey which is used in many elements) should be defined outside the palette?


lucienbertin:
> should text colors (standard grey, and light grey which is used in many elements) should be defined outside the palette?

maybe. as always my motto is

> if it makes no sesnse to have `button mod-xxx` like `mod-primary` then it shouldnt be in palettes

you can still have 

```scss
$palettes: (
  "primary": ( color: pink, text: white ),
  "secondary": ( color: green, text: white ),
  ...
),
$colors: ( 
  "grey": #cbcbcb,
  "light": #efefef,
  ....
)
```

PR #49, opened 2017-05-30
Scss.color.vs.palette

lucienbertin:
## Breaking
 - renamed `$luTheme` into `$theme` - needs refactor if you override it
 - renamed `palette` in `$theme` into `palettes` to be more coherent
 - removed function `_palette(keys)`, use `_color(keys)` instead

-------------

differentiate between colors and palettes, palettes should only be used to decline elements
refactor so `_palette(xxx.color)` isn't used anymore
for `button`, `label` and `callout`, loop on palettes to autogenerate classes declined from palettes


packages/scss/src/components/_button.scss
```diff
@@ -56,84 +56,42 @@
...
-
-	&.mod-success {
-		background: _palette("success.color");
-		color: _palette("success.text");
-
-		&:active {
-			background: darken(_palette("success.color"), 5%);
+}
+// standard coloring
+.button:not(.mod-flat):not(.mod-wired):not(.mod-filling) {
```
jeremie-lucca: We should not have mod references outside the mod itself (or in state in some exceptions)

packages/scss/src/components/_callout.scss
```diff
@@ -21,22 +21,16 @@
...
+@mixin calloutDefaultPalette($palette) {
+	&::before {
+		background: _get($palette, "color");
 	}
-
-	&.mod-error {
-		&::before {
-			background: _palette("error.color");
+}
+.callout:not(.mod-filled) {
```
jeremie-lucca: Same note as for buttons

packages/scss/src/components/_nav-side.scss
```diff
@@ -62,7 +63,7 @@
 
 .navSide-link {
 	&.is-active {
-		background: #404040;
+		background: #404040; // :'(
```
jeremie-lucca: Indeed, navSide needs a refactor. I'm waiting to have the new version form UX team to start it.

packages/scss/demo/components/callouts.html
```diff
@@ -12,24 +12,38 @@ <h2 class="demo-divider">Callout</h2>
 </code>
 
 	<!-- Mods -->
-	<div class="callout mod-success">
+	<div class="callout palette-primary">
```
jeremie-lucca: Do we need to provide a primary / secondary component in the demo?
I fell like it's not an actual use of callouts.

PR #53, opened 2017-06-02
Scss.fix layout

jeremie-lucca:


packages/scss/demo/templates/form.html
```diff
@@ -10,51 +10,51 @@
...
 	<body>
-		<div class="global u-paddingTopBiggest">
-			<div class="container mod-sm">
-				<div class="card">
-					<form class="form">
-						<div class="form-group">
-							<div class="input is-filled">
-								<input class="input-field" type="text" value="input value">
-								<label class="input-label">Input label</label>
+		<main class="main mod-noAside u-paddingTopBiggest">
```
sanlucca: mod-noAside might be handled through sibling selector instead.

jeremie-lucca: That could be a good idea.
As the layout can contain various major elements, I'm not sure we'll be able to do it, but we can have a try.

lucienbertin:
@sanlucca is it ok for you ?

packages/scss/src/components/_main.scss
```diff
@@ -1,33 +1,25 @@
...
-	padding-left: _theme("component.navSide.width");
-}
-
 .main-content {
 	padding: 4rem 0;
 }
 
-.main-filters {
-	margin-bottom: 3rem;
+.navSide + .main-content {
```
sanlucca: We might need to use ".navSide ~ .main-content" instead of ".navSide + .main-content" to cope with the case of div being put between the two.

jeremie-lucca: Good idea!

PR #55, opened 2017-06-05
Rework.override

lucienbertin:
- rework of the theme override process
 - BREAKING - reorganization of the theme and commenting all the stuff thats not compiling anymore
 - rewrite button styles to use it as example of how to write an component


packages/scss/src/components/_button.scss
```diff
@@ -227,40 +213,43 @@
...
-				font-size: 1rem;
-			}
-		}
-
-		&.mod-small {
-			&::after {
-				font-size: 1.2rem;
-			}
+// sizing for states
+.button.is-loading {
```
sanlucca: Shouldn't state sizing be including in the sizing loop ? (because it is the sizing of the button that changes how the state is displayed, not the other way around, even though it doesn't change anything in terms of css)

packages/scss/src/utilities/_text.scss
```diff
@@ -16,14 +16,14 @@
...
 // ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
-@mixin textPalette($palette) {
+@mixin textColoring($palette) {
 	color: _get($palette, "color") !important;
 }
 @each $name, $palette in _palettes() {
-	.u-text-#{$name} {
-		@include textPalette($palette);
+	.u-text#{str-capitalize($name)} {
+		@include textColoring($palette);
```
sanlucca: Couldn't we use the content of the mixin rather than having a 1-line mixin used 1 time ?

packages/scss/src/components/_button.scss
jeremie-lucca: In the last refacto, I simplified it by putting the sizing loop inside each mod. But indeed, having a big sizing loop which include each mod might be a good idea.

lucienbertin:
@sanlucca if its good for you, its good for me

PR #57, opened 2017-06-06
Scss.fix grid name

jeremie-lucca:
As these classes are still not used, I didn't tagged it as "breaking changes".


lucienbertin:
you might wanna autogenerate this i think

PR #58, opened 2017-06-07
Created media query mixins

sanlucca:
Created three media query mixins : 
1) media_smaller_than(breakpoint)
2) media_larger_than(breakpoint)
3) media_between(breakpoint1, breakpoint2)

They throw warnings if the breakpoint used isn't defined

### DIFF
--- a/packages/scss/src/mixins/_responsiveness.scss
+++ b/packages/scss/src/mixins/_responsiveness.scss
@@ -1,0 +1,35 @@
+@mixin media_smaller_than($breakpoint) {
+	@if map-has-key(map-get($theme,"breakpoints"), $breakpoint) {
+		@media (max-width: #{_theme("breakpoints.#{$breakpoint}.breakAt")}) {
+			@content;
+		}
+	}
+	@else {
+		@warn "We couldn't find a value for the breakpoint `#{$breakpoint}`."
+		+ " Check if it is defined in the `breakpoints` map";
+	}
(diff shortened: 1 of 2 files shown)


packages/scss/src/mixins/_responsiveness.scss
```diff
@@ -0,0 +1,35 @@
+@mixin media_smaller_than($breakpoint) {
+	@if map-has-key(map-get($theme,"breakpoints"), $breakpoint) {
```
lucienbertin: `_theme("brerakpoints")` is the same as `map-get($theme,"breakpoints")`

sanlucca: Yes, but if the key doesn't exist, it throws a general error (stating that the key $breakpoint doesn't exist) without specifying the fact that it is a breakpoint that doesn't exist (error handled by these mixins)

lucienbertin:
```
$stuff: _safeGet($theme, "breakpoints." + $breakpoint)
@if $stuff
  ...
```

lucienbertin: also i said that this
```
map-has-key(_theme("breakpoints"), $breakpoint)
```
is equivalent to that
```
map-has-key(map-get($theme,"breakpoints"), $breakpoint)
```

PR #60, opened 2017-06-13
CSS Animations & demo functions

sanlucca:
Add css animations as utilities.
A demo.js has been added to the demo to provide simple interaction functions and better handling of repetitive tasks (for icons)


packages/scss/src/theming/_commons.scss
```diff
@@ -1,4 +1,8 @@
 $commons: (
+	"animations": (
+		"duration": 300ms,
```
lucienbertin: it may be relevant to have 2 or 3 animation speed, `fast`, `default/normal`, `slow`

PR #65, opened 2017-06-28
Scss font-face imports

bmoriniere:
- Scss : possibility to override icons font url
- Icons : ensure @font-face is generated at root of css

### DIFF
--- a/packages/icons/src/_font.scss
+++ b/packages/icons/src/_font.scss
@@ -1,24 +1,26 @@
 $font-path: "lucca-icons" !default;
-@font-face {
-	font-family: "Lucca icons";
-	font-style: normal;
-	font-weight: 400;
-	src:  url('#{$font-path}.eot');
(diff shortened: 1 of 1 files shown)


packages/scss/src/_icons.scss
```diff
@@ -1,4 +1,4 @@
-$font-path: _theme("icons.font-path");
+$font-path: _theme("icons.font-path") !default;
```
sanlucca: As you should override the path in the variable map rather than defining it as a variable in your scss files, the "!default" shouldn't be necessary here.

bmoriniere:
Ok sorry, my bad ! I could not get the theme overriding to work properly, but I finally found out how to do it.
I removed that commit, but I left the commit about loading font-face at root. Without that if we include lucca-front inside our app tag (so its styles does not affect outside elements), the @font-face is nested and so is not valid.

PR #67, opened 2017-06-29
[enh] action icon button as component

jeremie-lucca:
- Action icon button as component (previously in list component)
- Fix icons sizes 
- Fix default icons size
- lucca-icon as default class

Breaking changes in for figgo's lists. Will be fixed after the release. 

image

### DIFF
--- a/packages/scss/src/components/_actionIcon.scss
+++ b/packages/scss/src/components/_actionIcon.scss
@@ -1,0 +1,20 @@
+.actionIcon {
+	background: transparent;
+	border: 0;
+	border-radius: 50%;
+	color: _color("text.default");
+	cursor: pointer;
+	height: 2rem;
+	margin-left: .1rem;
+	transition: all _theme("commons.animations.durations.fast") ease;
+	width: 2rem;
+
+	&:hover {
+		background: _component("actionIcon.hover.background");
+		color: darken(_color("text.default"), 25%);
+	}
+
+	.lucca-icon, .#{_theme("icons.prefix")} {
+		font-size: .9rem;
+	}
+}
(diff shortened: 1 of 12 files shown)


packages/scss/src/components/_actionIcon.scss
```diff
@@ -0,0 +1,20 @@
+.actionIcon {
```
sanlucca: You should use action-icon as we don't have any camelCase syntax elsewhere in the framework

packages/scss/src/components/_actionIcon.scss
```diff
@@ -0,0 +1,20 @@
...
+	background: transparent;
+	border: 0;
+	border-radius: 50%;
+	color: _color("text.default");
+	cursor: pointer;
+	height: 2rem;
```
sanlucca: Shouldn't it be em rather than em here? These type of buttons are really context sensitive and needs to adapt to the size of their context

packages/scss/src/components/_actionIcon.scss
jeremie-lucca: Multiple components use camelCase, inside a component, the - is used to separate parents and childs (e.g. .navSide-link)

packages/scss/src/components/_actionIcon.scss
jeremie-lucca: Hum... I feel like it could lead to unwanted size modifications if the button is linked to his parent size. So I'd rather add a size option to keep the component encapsuled.

sanlucca: I do think these buttons are meant to be linked to its parent size. Size options should exist to make them larger or smaller relative to their parent size.

jeremie-lucca: Here is a test with em : 
image
If I reduce it's size, default will be too small. Si how should I deal with sizing, then ? :/

I'm really attached to have encapsuled component. Furthermore, even basic icons have a .size-xxx option. So why especially this component should be parent relative?

sanlucca: ok

PR #72, opened 2017-07-05
Scss.icon actions disabled state

jeremie-lucca:
Alpha on hover to be more efficient on all backgrounds (especially on background's light grey)

### DIFF
--- a/packages/scss/src/components/_actionIcon.scss
+++ b/packages/scss/src/components/_actionIcon.scss
@@ -10,11 +10,24 @@
 	width: 2rem;
 
 	&:hover {
-		background: _component("actionIcon.hover.background");
-		color: darken(_color("text.default"), 25%);
+		&:not([disabled]) {
+			background: _component("actionIcon.hover.background");
+			color: darken(_color("text.default"), 25%);
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


packages/scss/src/components/_actionIcon.scss
```diff
@@ -10,11 +10,24 @@
 	width: 2rem;
 
 	&:hover {
-		background: _component("actionIcon.hover.background");
-		color: darken(_color("text.default"), 25%);
+		&:not([disabled]) {
```
sanlucca: Isn't possible to remove this :not([disabled]) by adding a pointer-events set to none when disabled ?

jeremie-lucca: I used a not:disabled to allow tooltip over disabled buttons.

sanlucca: Yes, good point. You should add a class "disabled" too, because all elements shouldn't have a "disabled" attribute

jeremie-lucca: Ok, I used the same code that we used for buttons. 
A class is-disabled has been added with a pointer event none.
I also fixed opacity which was not linked to common.disabled.

PR #74, opened 2017-07-05
Scss.input.larger

jeremie-lucca:


packages/scss/src/theming/components/_input.theme.scss
```diff
@@ -0,0 +1,11 @@
+$input: (
+  sizes: (
+    shortest: 3rem,
+    shorter: 6rem,
+    short: 9rem,
```
sanlucca: You should had a "default" size set to 13rem here, just to make all sizes accessible in a single file

PR #78, opened 2017-07-31
[enh] fix sizes bug + color error + warning error

jeremie-lucca:


packages/scss/src/components/_input.scss
```diff
@@ -224,6 +232,10 @@
 
 	&.is-error {
 		@include formError();
```
sanlucca: formError shouldn't be a mixin but be turned into an @extend as it will render into a more compact css.

packages/scss/src/components/_input.scss
```diff
@@ -224,6 +232,10 @@
 
 	&.is-error {
```
sanlucca: Isn't "&is-error, &is-error:focus" { ... } more clear ?

PR #80, opened 2017-07-31
Scss.reset spacing

jeremie-lucca:


packages/scss/src/utilities/_spacing.scss
```diff
@@ -7,218 +7,47 @@
...
-
-.u-paddingBottomStandard {
-	padding-bottom: _theme("spacings.standard") !important;
-}
+@each $name, $spacings in _theme("spacings") {
 
-.u-paddingBottomBig {
-	padding-bottom: _theme("spacings.big") !important;
-}
+	&.u-marginBottom#{str-capitalize($name)} {
```
sanlucca: What are the `&` used for here ?

jeremie-lucca: typo

PR #87, opened 2017-08-29
@lucca-front/ng DateRangePicker

versinia:
In its first version, this date range picker will allow to select a date range from a pre-configured list (no custom calendar), and is only compatible with `moment.js`.

Uses : 
 - moment
 - MdSelect
 - MdDatePicker

As this is the first component of the ng package, this PR also introduces [lucca-front translate feature](https://github.com/LuccaSA/lucca-front/pull/86).


packages/ng/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "ng": "ng",
-    "start": "gulp generatedoc && ng serve -a=demo",
+    "start": "gulp generatedoc && ng serve -a=demo --port 4242",
```
lucienbertin: why ?

packages/ng/src/app/date-range-picker/date-range-picker.models.ts
```diff
@@ -0,0 +1,11 @@
+import * as moment from 'moment';
+
+export interface DateRangeSelectChoice {
```
lucienbertin: start interface with `I`, `IDateRangeSelectChoice`

packages/ng/src/app/date-range-picker/date-range-picker.models.ts
```diff
@@ -0,0 +1,11 @@
...
+	range: DateRange;
+	label: string;
+}
+
+export interface DateRange {
```
lucienbertin: idem

packages/ng/src/app/date-range-picker/custom-range-picker/custom-range-picker.component.html
```diff
@@ -0,0 +1,38 @@
+<div class="custom-range-picker-inputs">
```
lucienbertin: call all this date range modal instead of custom range wha'eve'

lucienbertin: by that i mean the drectpory + the differnet fisles

packages/ng/src/app/date-range-picker/date-range-picker.models.ts
```diff
@@ -0,0 +1,11 @@
+import * as moment from 'moment';
```
lucienbertin: `date-range-picker.model.ts` - model singular plz

packages/ng/src/app/date-range-picker/date-range-picker.models.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface DateRangeSelectChoice {
+	range: DateRange;
+	label: string;
+}
+
+export interface DateRange {
+	dateMin: moment.Moment;
+	dateMax: moment.Moment;
+}
```
lucienbertin: also, use `start` and `end`, you're gonna need `min` and `max` when it will support some min and max

packages/ng/package.json
versinia: This was commited by mistake. I work on another project using :4200. Sorry about that ☘️

packages/ng/src/app/date-range-picker/date-range-modal/date-range-modal.component.html
```diff
@@ -1,7 +1,6 @@
 <div class="date-range-modal-inputs">
 	<div class="input mod-inline">
 		<input class="input-field"
-					 [ngClass]="{'is-filled': !!start}"
```
versinia: @sanlucca This is used to keep label on top position if the input is filled, and switch to error style if it's empty, correct ? In this case we'd like the former to happen, but not the latter.

sanlucca: The "is-error" state and the "is-filled" state aren't correlated. It can be both filled and error.

packages/ng/src/app/shared/moment/moment.date-adapter.ts
```diff
@@ -0,0 +1,123 @@
...
+	getISODateString(date: moment.Moment): string {
+		return date.format();
+	}
+
+	isDateInstance(obj: any): boolean {
+		return moment.isMoment(obj);
+	}
+
+	isValid(date: moment.Moment): boolean {
+		return date.isValid();
```
sanlucca: date.isValid considers an empty string as invalid. We need a way to consider an empty string as a valid date.

versinia: @sanlucca @lucienbertin the problem resides on our choice not to use `md-form-field` on datepicker-attached inputs (it greatly simplifies material-inejcted DOMS). 

What we could do is make our `DateAdapter.parse()` method return `null` for an empty string input (not an absurd choice, the default Material DateAdapter does just that), and then make the `DateAdapter.isValid()` method return true for a `null` input.

PR #88, opened 2017-09-01
ng.datepicker style

sanlucca:
Style for the datepicker/daterangepicker.
The plugin/ngMaterial folder contains the necessary style to use Material2 components and overrides the necessary classes.

### DIFF
--- a/packages/ng/src/plugins/ngMaterial/_general.scss
+++ b/packages/ng/src/plugins/ngMaterial/_general.scss
@@ -1,0 +1,13 @@
+@import
+	'~@angular/material/theming', // Basic file for custom theme
+	'mixins',
+	'overlay',
+	'buttons',
+	'datepicker',
+	'select',
+	'dialog';
+
+// CHANGE DEFAULT FONTS TO LUCCA-FRONT FONTS
+@include angular-material-typography(mat-typography-config(
+	$font-family: _theme("commons.font.family")
+));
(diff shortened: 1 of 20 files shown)


packages/ng/src/plugins/ngMaterial/_general.scss
```diff
@@ -0,0 +1,13 @@
+@import
+	'~@angular/material/theming', // Basic file for custom theme
```
lucienbertin: is this mandatory ?

packages/ng/src/style/_components.scss
```diff
@@ -1,3 +1,4 @@
 @import
 
-  'components/input';
+  'components/input',
+  '../plugins/ngMaterial/general';
```
lucienbertin: why do you import this here ?, in `_components` ? we may distribute components that dont need to override ng material

packages/ng/src/plugins/ngMaterial/_general.scss
sanlucca: it gives access to the `angular-material-typography` mixin which allow us to easily override fonts all over material components

packages/ng/src/style/_components.scss
sanlucca: I put it there by default and forgot to double check afterwards. I have to try if we can import it directly into each component scss when needed, but I'm afraid some overrides (overlay and anything that end up into the overlay zone) might need to be imported globally.

PR #91, opened 2017-09-05
Scss.filter layout

jeremie-lucca:


packages/scss/src/components/_header.scss
```diff
@@ -9,3 +9,12 @@
...
 	z-index: 1;
 }
+
+.header-left {
+	display: flex;
+}
+
+.header-right {
+	display: flex;
+	margin-left: auto;
```
sanlucca: If we go this path, you should remove the "justify-content: space-between" from the `header` class

packages/scss/src/components/_header.scss
```diff
@@ -9,3 +9,12 @@
 	position: relative;
 	z-index: 1;
 }
+
+.header-left {
```
sanlucca: It should be `header-content-left/right` otherwise we might think it is a header that lives on the left/right side

packages/scss/src/components/_header.scss
jeremie-lucca: woops :)

PR #92, opened 2017-09-05
[enh] switch disabled color

jeremie-lucca:


packages/scss/src/components/inputs/switch/_switch.scss
```diff
@@ -108,3 +80,48 @@
...
+				}
+
+				&::after {
+					background: _get($palette, "color");
+				}
+			}
+
+			// Palette > Disabled switch 
+			&[disabled], &[readonly]{
+				&:checked {
```
sanlucca: You're already selecting `.switch-input:checked`, this one doesn't seem necessary

PR #94, opened 2017-09-08
Ng.user tile

versinia:
Tile for user display. By default, the header displays the user's `jobTitle` but is customizable. A footer can be added via `ng-content`. 

This branch is based on `ng.datepicker` for convenience, but will be rebased and merged onto master after `ng.datepicker` is merged.  This PR is for review only.


versinia:
Closing PR as it is for review only.

PR #96, opened 2017-09-12
Scss.menu.theming

jeremie-lucca:
oops @ firsts commits naming

### DIFF
--- a/packages/scss/src/components/_menu.scss
+++ b/packages/scss/src/components/_menu.scss
@@ -6,15 +6,14 @@
 
 .menu-link {
 	color: _color("text.default");
-	margin: 0 1.2rem;
+	margin: _component("menu.margin");
 	opacity: .75;
-	padding: 1.5rem 0;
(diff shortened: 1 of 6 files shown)


packages/scss/src/components/_menu.scss
```diff
@@ -6,15 +6,14 @@
...
 	color: _color("text.default");
-	margin: 0 1.2rem;
+	margin: _component("menu.margin");
 	opacity: .75;
-	padding: 1.5rem 0;
+	padding: _component("menu.padding");
 	position: relative;
 	text-decoration: none;
-	transition: opacity 250ms ease;
+	transition: opacity _theme("commons.animations.durations.standard") ease;
```
lucienbertin: you can use `_component("menu.animations.durations.standard")` here, it will fallback to `commons.animations.durations.standard` if it doesnt find `animations.durations.standard` under `component.menu` and still allows you to override said variable just for this component

PR #98, opened 2017-09-19
[scss][enh] list item hovering

jeremie-lucca:
image
image

Small breaking on mod / link position

### DIFF
--- a/packages/scss/src/theming/components/_list.theme.scss
+++ b/packages/scss/src/theming/components/_list.theme.scss
@@ -1,5 +1,8 @@
 $list: (
   margin-bottom: 1rem,
   padding: 1rem 1.5rem,
+  hover: (
+    background: #EEEEEE
+  )
 );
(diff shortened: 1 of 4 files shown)


packages/scss/src/theming/components/_list.theme.scss
```diff
@@ -1,5 +1,8 @@
 $list: (
   margin-bottom: 1rem,
   padding: 1rem 1.5rem,
+  hover: (
+    background: #EEEEEE
```
sanlucca: I wonder if we shouldn't add in the $colors map a "background" map including default white and variations of greys

jeremie-lucca: I was thinking to use a common.hovering.background 🤔

sanlucca: Will be done in another PR

PR #99, opened 2017-09-20
[FIX] Icons were moving text in navSide-item-link

sanlucca:


packages/scss/src/theming/components/_nav-side.theme.scss
```diff
@@ -8,6 +8,8 @@ $navSide: (
 		text: _theme("palettes.primary.text"),
 	),
 	"offset-top": 0px,
+	"main-font-size": 1rem,
+	"sub-font-size": .9em,
```
jeremie-lucca: May we would like to use _theme("sizes.standard.font-size") & _theme("sizes.small.font-size") there ?

PR #106, opened 2017-09-21
Changes how the navSide is themed

sanlucca:
Removed any darken/lighten modificator from the style. Now every color can be tweaked in the compact-palette and the fullwidth-palette.


packages/scss/src/components/_nav-side.scss
```diff
@@ -66,9 +66,9 @@ nav.navSide-scrollWrapper {
 }
 
 .navSide-item-alert {
-	background: _component("navSide.accent-palette.color");
+	background: _component("navSide.fullwidth-palette.alert-color");
 	border-radius: .5rem;
-	color: _component("navSide.accent-palette.text"); // lulu - i dont like it
+	color: _component("navSide.fullwidth-palette.alert-text"); // lulu - i dont like it
```
jeremie-lucca: Can we remove this comment ? 😁

PR #112, opened 2017-09-26
forms & input mod-framed

sanlucca:
Add mod-framed to form & input. It handles `.form-group` differently and adds `.form-group-line` , `.form-group-title`  and `.input-suffix`.
chrome_2017-09-26_10-20-45


packages/scss/src/components/inputs/checkbox/_checkbox.framed.scss
```diff
@@ -0,0 +1,11 @@
+%checkbox-framed {
+	@include fakeBorder(1px, _theme("commons.divider.color"));
+
+	display: flex;
+	flex-wrap: wrap;
+	padding: _component("input.framed.top-padding") _component("input.framed.side-padding") _component("input.framed.bottom-padding");
```
jeremie-lucca: Why don't we use a sigle "x x x x" padding value?

sanlucca: You mean as a variable ? Because we need the individual values elsewhere.

packages/scss/src/components/inputs/checkbox/_checkbox.framed.scss
```diff
@@ -0,0 +1,11 @@
...
+	width: 100%;
+	.checkbox {
```
jeremie-lucca: Each component should be encapsuled

packages/scss/src/components/inputs/input/_input.framed.scss
```diff
@@ -0,0 +1,142 @@
...
+	@include inputColoring(_component("input.framed.default-palette"));
+
+	// palette colors
+	@each $name, $palette in _palettes() {
+		&.palette-#{$name} {
+			@include inputColoring($palette);
+		}
+	}
+	// MOD
+	&.has-suffix {
```
jeremie-lucca: Should be mod-withSuffix

packages/scss/src/components/inputs/checkbox/_checkbox.framed.scss
```diff
@@ -0,0 +1,11 @@
+%checkbox-framed {
+	@include fakeBorder(1px, _theme("commons.divider.color"));
+
+	display: flex;
```
jeremie-lucca: What about removing display flex and checkbox's right margin to use checkbox.mod-input which is implemented since 0.0.17 ? I don't know if it will be a use case but it would allow to use stacked and inline checkbox / radio / etc.

packages/scss/demo/components/forms/forms-framed.html
```diff
@@ -0,0 +1,186 @@
...
+				<label class="input-label">Input label</label>
+			</div>
+			<div class="input mod-framed has-suffix">
+				<input class="input-field" type="text" placeholder="Money Money">
+				<label class="input-label">Input label</label>
+				<div class="input-suffix">
+					€
+				</div>
+			</div>
+			<div class="input textarea mod-framed has-suffix">
```
jeremie-lucca: input textarea < Double component?

packages/scss/src/components/inputs/input/_input.framed.scss
```diff
@@ -0,0 +1,142 @@
...
+			}
+		}
+
+		&.is-error {
+			background-color: transparentize(_color("error"), .9);
+			&:focus, &:hover {
+				@include fakeBorder(1px, _color("error"));
+			}
+
+			~ label, ~ .input-suffix {
```
jeremie-lucca: Should be a class name

packages/scss/src/components/inputs/input/_input.framed.scss
```diff
@@ -0,0 +1,142 @@
...
+		}
+	}
+	// STATUS
+
+	.input-field {
+		&[disabled], &[readonly] {
+			background-color: transparentize(_color("text.placeholder"), .8);
+			background-image: none;
+			opacity: .6;
+			& + label {
```
jeremie-lucca: Should be a class name

packages/scss/demo/components/forms/forms-framed.html
```diff
@@ -0,0 +1,186 @@
...
+		In this mod <code class="code">.form-group</code> behave differently and you have access to a new
+		DOM level called <code class="code">.form-group-line</code>.
+	</p>
+	<p>
+		<code class="code">form-group</code> represents a whole section of your form.
+		It can have a <code class="code">.form-group-title</code> to name your section
+		It can contain <code class="code">.form-group-line</code> and <code class="code">.input.mod-framed</code>
+	</p>
+	<p>
+		In a <code class="code">form-group-line</code> you can use <code class="code">.grid-*</code> classes onto your <code class="code">.input</code> to size
```
jeremie-lucca: grid- are supposed to have a grid parent, so I'm not sure that it's a clean way to do it.
Furthermore, as grids are not used here as containers, it may be possible to break framed form by updating grid component.
So what about using form framed relative mod classes? (Something like mod-framed1/2/3 with flex: 1/2/3 attributes).

sanlucca: I didn't want to duplicate the grid system elsewhere that's why I went this way. But has it is a lightweight system with only 2-3 options, it might be worth a shot.

packages/scss/demo/components/forms/forms-framed.html
sanlucca: Yes, it avoids style duplication between input.mod-framed and textarea.mod-framed. It may lead to a transformation of textarea as a mod-textarea of input later

packages/scss/src/components/inputs/checkbox/_checkbox.framed.scss
sanlucca: I'd prefer to had a mod-stacked to the radio/checkbox-group that would change the flex-direction into column as it is more aligned with the philosophy of this design

packages/scss/src/components/inputs/input/_input.framed.scss
sanlucca: I'll add is-disabled and is-readonly to every input

packages/scss/src/components/inputs/input/_input.framed.scss
sanlucca: What should be ?

packages/scss/src/components/inputs/checkbox/_checkbox.framed.scss
jeremie-lucca: ok

PR #113, opened 2017-09-26
Upgrading ng package to material 2.0.0-beta11

versinia:
[material2.0.0-beta 11 has landed](https://github.com/angular/material2/releases/tag/2.0.0-beta.11), and brings some change that will be breaking (effective from next release), as well as an official adapter for Moment dates.


versinia:
The Moment adapter provided by Material is not accessible by import (I'm not sure why 🤔), so I simply replaced our custom implementation with their source code.

PR #116, opened 2017-09-28
Ng.formly

lucienbertin:
- [x] all dat style
 - [x] support for inputs like select, autocomplete, textarea
 - [x] directive to flag empty input, since `ng-empty` doesnt exist anymore :'(
 - [x] PR #112 - support for input framed


packages/ng/src/app/formly/wrappers/error.ts
```diff
@@ -0,0 +1,56 @@
+import { Component, Input, ViewChild, ViewContainerRef } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { FormlyFieldConfig, FieldWrapper, FormlyConfig } from 'ng-formly';
+import * as _ from 'underscore';
```
GuillaumeNury: Why underscore and not lodash ?
Anyway, prefer `import { chain } from 'underscore'` as it avoid including the whole library

PR #123, opened 2017-10-05
Username.pipe

pierop:
Should merge https://github.com/LuccaSA/lucca-front/pull/122 first
Add pipe to display a user name according to specific formats


packages/ng/src/app/user/user-display.pipe.ts
```diff
@@ -0,0 +1,43 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { IUser } from './user.model';
+
+/**
+ * Displays a user name according to specified format. Supported formats: f for first name, F for first initial, l for last name, L for last initial.
```
lucienbertin: i suppose it doesn't work super well for people like _Jean Francois Garreau_

PR #130, opened 2017-10-17
[enh] input/textarea => textfield

jeremie-lucca:


packages/scss/src/components/inputs/textfield/_textfield.framed.scss
```diff
@@ -0,0 +1,186 @@
...
+		@include inputColoring(_component("textfield.compact.default-palette"));
+
+		// palette colors
+		@each $name, $palette in _palettes() {
+			&.palette-#{$name} {
+				@include inputColoring($palette);
+			}
+		}
+
+		// MODS
```
sanlucca: We need to specify a style for `.textfield-suffix` here (.mod-framed.mod-multiline) that pushes it to the bottom right corner.
Maybe :
top: auto;
bottom: XXem;

PR #135, opened 2017-10-20
formly [add] class is-focued and is-error on the layout

lucienbertin:
adds classes `is-focused` and `is-error` to layout


sanlucca:
And what about the is-required state ?

PR #139, opened 2017-10-27
Scss.navside.revision

jeremie-lucca:


packages/scss/src/theming/components/_nav-side.theme.scss
```diff
@@ -1,26 +1,27 @@
 $navSide: (
 
 	"fullwidth-palette": (
-		bg-color: _color("primary"),
-		text: _theme("palettes.primary.text"),
-		selected-bg: darken(_color("primary"), 5%),
-		selected-text: _theme("palettes.primary.text"),
-		alert-color: darken(_color("primary"), 5%),
+		bg-color: _color("text.default"),
+		text: white,
```
sanlucca: I would prefer to keep the default palette to be linked to the palettes. For both fullwidth and compact. The issue here is more about the palettes used in the demo. It shouldn't affect these default variables

PR #173, opened 2017-11-29
[SCSS] Mobile support for navSide

sanlucca:
Adds mobile support for navSide

### DIFF
--- a/packages/scss/demo/templates/fullmenu.html
+++ b/packages/scss/demo/templates/fullmenu.html
@@ -11,8 +11,14 @@
 	</head>
 	<body>
 		<main class="main">
-			<aside class="navSide">
+			<aside class="navSide" id="navSide">	
 				<nav class="navSide-scrollWrapper">
+					<div class="navSide-item navSide-mobile-toggle" onclick="w3.toggleClass('#navSide', 'is-open')">
+						<a href="#" class="navSide-item-link">
+							<span class="navSide-item-link-title">Menu</span>
+							<i class="lucca-icon navSide-item-arrow">arrow_south_thin</i>
+						</a>
+					</div>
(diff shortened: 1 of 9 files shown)


packages/scss/demo/templates/fullmenu.html
```diff
@@ -11,8 +11,14 @@
 	</head>
 	<body>
 		<main class="main">
-			<aside class="navSide">
+			<aside class="navSide" id="navSide">	
 				<nav class="navSide-scrollWrapper">
+					<div class="navSide-item navSide-mobile-toggle" onclick="w3.toggleClass('#navSide', 'is-open')">
```
jeremie-lucca: Should have only one class, with only one child (navSide-toggle or navSide-mobileToggle for example)

packages/scss/src/components/_main.scss
```diff
@@ -14,3 +14,16 @@
...
 }
+
+@include media_smaller_than("sm") {
+	.main-content {
+		margin-left: 0 !important;
+		overflow-x: hidden;
+	}
+
+	// MOD TO BE ADDED TO .MAIN WHEN THE MENU IS TOGGLED
+	.main.is-noScroll {
```
jeremie-lucca: is-unscrollable ? :p

packages/scss/src/components/_nav-side.scss
```diff
@@ -266,15 +216,109 @@
...
+@include media_smaller_than("sm") {
+	.navSide {
+		position: relative;
+		width: 100%;
+
+		&.mod-withBanner {
+			margin-top: _theme("commons.banner-height");
+		}
+	}
+
```
jeremie-lucca: What about a slide down by default on navSide-item?

PR #174, opened 2017-11-29
Ng.demo.redirect

lucienbertin:
for the directives api-picker and user-picker, the demo needs to be able to redirect xhr requests to a lucca website. so this pr adds the redirectModule in the demo shared module to allow users to connect to one of their local website or a prod website (#totesSafe)

the component looks like this

image

image


lucienbertin:
you can test it on ttab [/api](http://localhost:4200/api), clicking on who am i calls `/api/v3/users/me`

PR #199, opened 2017-12-18
[scss] checkbox / radio buttons

jeremie-lucca:


packages/scss/src/components/_button.scss
```diff
@@ -297,15 +296,56 @@
...
+	}
+
+	&:last-child {
+		.button {
+			border-radius: 0 _component("button.border.radius") _component("button.border.radius") 0;
+		}
+	}
+}
+
+.button-group-radio {
```
sanlucca: We could use .button-group-input and have a single class for both checkbox / radio

packages/scss/src/theming/components/_button.theme.scss
```diff
@@ -30,6 +30,9 @@ $button: (
 		color: black,
 	),
 
+	radio: _theme("palettes.primary"),
```
sanlucca: If we use a single class like button-group-input, we would use only one variable

packages/scss/src/components/_button.scss
```diff
@@ -297,15 +296,56 @@
 		position: relative;
 
 		&:first-child {
```
sanlucca: Maybe we could have these :first-child and :last-child moved to a `> .button` selector, to avoid having this style bleeding over the .button-group-item

packages/scss/src/components/_button.scss
```diff
@@ -297,15 +296,56 @@
...
 
 		&:hover, &:focus {
 			z-index: 1;
 		}
 	}
 }
+
+.button-group-item {
+
+	.button {
```
sanlucca: if we use the `> .button` selector above, we can get rid of this styling

PR #208, opened 2018-01-10
[enh] inline validation

jeremie-lucca:
Valid & Invalid icons don't match. It will need an update after next icons update.

### DIFF
--- a/packages/scss/src/components/inputs/textfield/_textfield.material.scss
+++ b/packages/scss/src/components/inputs/textfield/_textfield.material.scss
@@ -121,6 +121,14 @@
 	// ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
 
 	// Inline validation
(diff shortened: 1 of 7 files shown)


packages/scss/src/components/inputs/textfield/_textfield.material.scss
```diff
@@ -120,8 +120,30 @@
 	// STATES
 	// ▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒▒
 
-	.textfield-input {
+	// Inline validation
+	&.is-valid {
+		&::before {
+			content: "";
```
sanlucca: Why removing the icon in material ?

jeremie-lucca: Yep, we talked about that with @Clotilde. The idea is to use only the bottom line to give succes / error informations.

sanlucca: Then what is the need for the other mods? Aren't borders enough in these cases ?

jeremie-lucca:
image

PR #213, opened 2018-01-25
[NG] lu-user-tile & lu-user-picture new size attribute

sanlucca:
Adds new attributes to lu-user-tile and lu-user-picture:
- `lu-user-picture` has a new attribute `size` to set it size in px
- `lu-user-tile` has a new attribute `imageSize` to set its user-picture size in px

### DIFF
--- a/packages/ng/demo/app/user/picture/basic/basic.html
+++ b/packages/ng/demo/app/user/picture/basic/basic.html
@@ -3,4 +3,4 @@
 <p>Without picture</p>
 <lu-user-picture [user]="jeanMich"></lu-user-picture>
 <p>With specified size</p>
-<lu-user-picture [user]="anais" size="20"></lu-user-picture>
+<lu-user-picture [user]="anais" size="50"></lu-user-picture>
--- a/packages/ng/demo/app/user/picture/basic/basic.html
+++ b/packages/ng/demo/app/user/picture/basic/basic.html
@@ -2,3 +2,5 @@
 <lu-user-picture [user]="anais"></lu-user-picture>
 <p>Without picture</p>
 <lu-user-picture [user]="jeanMich"></lu-user-picture>
+<p>With specified size</p>
+<lu-user-picture [user]="anais" size="20"></lu-user-picture>
--- a/packages/ng/demo/app/user/picture/basic/basic.html
+++ b/packages/ng/demo/app/user/picture/basic/basic.html
@@ -2,5 +2,11 @@
 <lu-user-picture [user]="anais"></lu-user-picture>
 <p>Without picture</p>
 <lu-user-picture [user]="jeanMich"></lu-user-picture>
-<p>With specified size</p>
-<lu-user-picture [user]="anais" size="50"></lu-user-picture>
+<p>Size mods</p>
+<lu-user-picture [user]="anais" class="mod-smallest"></lu-user-picture>
+<lu-user-picture [user]="anais" class="mod-smaller"></lu-user-picture>
+<lu-user-picture [user]="anais" class="mod-small"></lu-user-picture>
+<lu-user-picture [user]="anais"></lu-user-picture>
+<lu-user-picture [user]="anais" class="mod-large"></lu-user-picture>
+<lu-user-picture [user]="anais" class="mod-larger"></lu-user-picture>
+<lu-user-picture [user]="anais" class="mod-largest"></lu-user-picture>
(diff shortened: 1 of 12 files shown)


packages/ng/demo/app/user/picture/basic/basic.html
```diff
@@ -2,3 +2,5 @@
 <lu-user-picture [user]="anais"></lu-user-picture>
 <p>Without picture</p>
 <lu-user-picture [user]="jeanMich"></lu-user-picture>
+<p>With specified size</p>
+<lu-user-picture [user]="anais" size="50"></lu-user-picture>
```
jefBinomed: As the final style is apply to `font-size` maybe `size` isn't the best name for this attribute. At first, i thought that `size` correspond to a kind of global ratio for the image

jefBinomed: With scss, as we do for "override_theme" can we open a kind of variable ? Or maybe it's the time to use css variables :p ? Because, it feels very strange to me to change the style of a component by passing values to attributes

packages/ng/demo/app/user/tile/basic/basic.component.html
```diff
@@ -3,6 +3,7 @@
 	<lu-user-tile class="user-tile" [user]="jeanPascal"></lu-user-tile>
 	<lu-user-tile class="user-tile" [user]="anais"></lu-user-tile>
 	<lu-user-tile class="user-tile" [user]="anais" [role]="'French actress'"></lu-user-tile>
+	<lu-user-tile class="user-tile" [user]="anais" [role]="'French actress'" imageSize="24"></lu-user-tile>
```
jefBinomed: Same remarq about sccs orverridable

packages/ng/src/app/user/picture/user-picture.component.ts
```diff
@@ -12,19 +12,28 @@ import { LuUserDisplayPipe } from '../display';
...
 	@Input() set user(user: IUser) {
 		this._user = user;
 		this.initials = this.displayPipe.transform(user, 'LF');
 		this.hasPicture = !!user.picture && !!user.picture.href;
+		if (this.size) {
+			this.style = { 'font-size': Math.round(this.size / 2.5) + 'px'};
+		}
 		if (this.hasPicture) {
-			this.style = { 'background-image': `url('${this._user.picture.href}')` };
+			this.style = { ...this.style, 'background-image': `url('${this._user.picture.href}')` };
```
jefBinomed: the fact you have to use `...this.style` means that you have in mind to add other attributes but others attributes means we are entering in a kind of black box with to many parameters... 

If we really want to go in this way, i prefer to play with a kind of CSS in JS like 

```javascript
{
'font-size':'Math.round(this.size / 2.5) + 'px',
'width' : ....
````

and this a directive for example : `[customStyle]="myObjectWithStyleOverride"`

File: packages/ng/demo/app/user/picture/basic/basic.html (hunk 1 above)
Comment (sanlucca) [reply packages/ng/demo/app/user/picture/basic/basic.html]: I use `font-size` to handle a good scaling of both font/padding and width/height. But it's called `size` because the value you give will actually be the final size in pixel of the picture. 
Regarding theme and variables, the thing as I used it in various interfaces, I realized how much customization is needed size-wise to adapt to each interfaces, therefore, variables wouldn't do the job as you might need various sizes in the same soft.
One solution could be to have mods (mod-small, mod-big, ...)

File: packages/ng/src/app/user/picture/user-picture.component.ts (hunk 3 above)
Comment (sanlucca) [reply packages/ng/src/app/user/picture/user-picture.component.ts]: I'm not sure I get your point here. I am using `...this.style` to 'concatenate' the sizing (that is handled by changing the font-size), with the picture image (that can either be a background-image or a background-color if no image is available). It's then used through ngStyle. But it doesn't change the fact that it has its own style defined through a scss file.

File: packages/ng/demo/app/user/tile/basic/basic.component.html [hunk 4]
```diff
@@ -1,8 +1,10 @@
 <div class="tiles">
 	<lu-user-tile class="user-tile" [user]="jeanMich"></lu-user-tile>
+
 	<lu-user-tile class="user-tile" [user]="jeanPascal"></lu-user-tile>
 	<lu-user-tile class="user-tile" [user]="anais"></lu-user-tile>
 	<lu-user-tile class="user-tile" [user]="anais" [role]="'French actress'"></lu-user-tile>
+	<lu-user-tile class="user-tile" [user]="anais" [role]="'French actress'" imageSize="24"></lu-user-tile>
```
jefBinomed: I think you forgot to clean the attribute for the demo ;)

PR #214, opened 2018-01-29
[ICONS] Adds classes and demo

sanlucca:
This PR adds classes to generate icons, based on the ligatures but in dash-case 
(eg. `<i class="lucca-icon">user_add</i>` becomes `<i class="lucca-icon user-add"></i>`).

It is added in a separate `main-with-classes.scss` file to avoid importing it when if not needed.

This `main-with-classes.scss` is the base file for the dist version, meaning that the minified CSS for lucca-icons will always have the icons classes available.


packages/icons/demo/demo.html
```diff
@@ -0,0 +1,44 @@
...
+		<div>
+			Lucca-icons can be used in two ways:
+			<ul>
+				<li>
+					using ligature: add the <code class="code">lucca-icon</code> class to your element and use the corresponding ligature (written here below each icon)
+					<span class="example">eg.: <code class="code">&lt;i class=&quot;lucca-icon&quot;&gt;user_roles&lt;/i&gt;</code></span>
+				</li>
+				<li>using classes: add the
+					<code class="code">lucca-icon</code> class and use the ligature in dash-case as a class. 
+					<span class="example">eg.:<code class="code">&lt;i class=&quot;lucca-icon user-roles&quot;&gt;&lt;/i&gt;</code></span>
```
jeremie-lucca: That could generate some classe conflicts in the future. Some icons names are very generic (menu, user, add, settings, calendar, edit, search, download, error, list, etc.) 

We should follow the kit's styleguide and use name as mods to encapsulate style : <i class="lucca-icon mod-user"></i>
Furthermore, it would be logic as "lucca-icon" is the generic part and "user", the modificator.

PR #218, opened 2018-02-08
[add] toasts

jeremie-lucca:
I set a keyframe animation which handle fadein and fadeout animations. 
When toasts will become a NG component, we'll be able to use NG animations.


packages/scss/src/components/_toasts.scss
```diff
@@ -0,0 +1,62 @@
...
+		opacity: .66;
+	}
+
+	&::after {
+		@include makeIcon("cross");
+		font-size: _theme("sizes.standard.font-size");
+	}
+}
+
+@keyframes toast {
```
sanlucca: You might want to use the keyframes in utilities/animations files.
Either way, if you prefer using one here, use translations instead of changing top for better performances.

jeremie-lucca: Indeed, keyframes was previously used for advanced animations (slide in / out) ans it was necessary to use a top positionning to affect others toasts positions. 
But as the toasts functionning changed, it's not needed anymore.

packages/scss/src/components/_toasts.scss
```diff
@@ -0,0 +1,62 @@
...
+	from {
+		transform: translateY(1rem);
+		opacity: 0;
+	}
+
+	to {
+		top: 0;
```
sanlucca: It's not needed anymore

PR #219, opened 2018-02-08
ng-build [mod] dont fix rxjs version

lucienbertin:
i have some transpilation problems cuz different libs want different versions of rxjs, by doing that i can tweak my lock file to stop having these issues

### DIFF
--- a/packages/ng/package.json
+++ b/packages/ng/package.json
@@ -46,7 +46,7 @@
     "@lucca-front/scss": "^0.2.2",
     "@ngx-translate/core": "^8.0.0",
     "core-js": "^2.4.1",
-    "rxjs": "5.5.2",
+    "rxjs": "^5.5.2",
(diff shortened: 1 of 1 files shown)


GuillaumeNury:
We could use RxJS as a peer dependancy ?

lucienbertin:
true but i'm not sure if peer dependencies are installed when running `npm i` or `yanr install`, and i believe it muist be installed for lf/ng to build

lucienbertin:
https://yarnpkg.com/en/docs/dependency-types

the line is fuzzy between deps and peer deps. i think it _should_ be a peer, but must be a direct dependency because it cant build without

also current packagers (npm, yarn) dont provide way to install the defined peer deps during `npm/yarn install`, meaning its way harder to plug automatic CI reliant on peer dependencies

https://github.com/yarnpkg/yarn/issues/1503

PR #221, opened 2018-02-13
Ng fix non selectable items 

jefBinomed:
Fix issue where options are not updated

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,7 +3,7 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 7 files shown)


lucienbertin:
?

jefBinomed:
@lucienbertin this is a fix for select because, if we load in an asynchronus way the options, the picker don't have the informations and the options couldn't be selectable. So I had to add the `notifyChange`.

All the lerna stuff is because I test it with directory team and i try to publish alpha version to test.

PR #227, opened 2018-02-19
LuTree & LuTreePicker components

bastienguillon:
Added two new components to display and select nodes in tree structures.

- `lu-tree`:
image

- `lu-tree-picker`, an `lu-tree` wrapped inside a dropdown
image

These components need a little makeover by @sanlucca to be used in production. The "(parent only)" label also needs to be translated, but I'm not sure if I can use ngx-translate on this repo.


packages/ng/src/app/tree-picker/picker/tree-picker.component.ts
```diff
@@ -0,0 +1,80 @@
...
+	private onTouchedCallback: (_: any) => void = () => { };
+	public registerOnTouched(methodToCall: (_: any) => void): void {
+		this.onTouchedCallback = methodToCall;
+	}
+
+	public ngOnInit() {
+		if (this.multiple == null) {
+			this.multiple = LuTreeComponent.DEFAULT_ALLOW_MULTIPLE;
+		}
+		if (this.value == null) {
```
jefBinomed: Why forcing a default value ? null or undefined isn't acceptable ?

packages/ng/src/app/tree-picker/picker/tree-picker.component.ts
```diff
@@ -0,0 +1,80 @@
...
+			this.value = this.multiple ? [] : {};
+		}
+		if (this.placeholder == null) {
```
jefBinomed: Same remarq about the previous about the placeholder, for me this code is not necessary

packages/ng/src/app/tree-picker/standalone-tree/tree-item/tree-item.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+			.compileComponents();
+	}));
+
+	beforeEach(() => {
+		fixture = TestBed.createComponent(LuTreeItemComponent);
+		component = fixture.componentInstance;
+		fixture.detectChanges();
+	});
+
+	it('should create', () => {
```
jefBinomed: You should integrate more tests.

packages/ng/src/app/tree-picker/standalone-tree/tree-item/tree-item.component.ts
```diff
@@ -0,0 +1,86 @@
...
+		this.tree.node.isSelected = !this.tree.node.isSelected;
+		this.messenger.notifyNodeChanged(new TreePickerMessage(this.tree.node, this.nodeUniqueId));
+	}
+
+	/**
+	 * Generates a GUID.
+	 * See: https://stackoverflow.com/a/2117523/2456022
+	 * TODO: export this method in a helper
+	 */
+	private guid(): string {
```
jefBinomed: Are you shure that this method couldn't generarte conflics of guid ? as it is a random method at the source of datas.... A simple counter like they do in angular material do the job to

packages/ng/src/app/tree-picker/standalone-tree/tree.class.ts
```diff
@@ -0,0 +1,67 @@
+export interface ITree {
+	node: ITreeNode;
+	children: ITree[];
+}
+
+export interface ITreeNode {
+	id: number;
```
jefBinomed: Why is there no `<T>`Value for a node ?

packages/ng/src/app/tree-picker/standalone-tree/tree.component.html
```diff
@@ -0,0 +1,5 @@
+<ul>
```
jefBinomed: could the tree-item be enough ? As it is alway reccursive

packages/ng/src/app/tree-picker/standalone-tree/tree.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+			.compileComponents();
+	}));
+
+	beforeEach(() => {
+		fixture = TestBed.createComponent(LuTreeComponent);
+		component = fixture.componentInstance;
+		fixture.detectChanges();
+	});
+
+	it('should create', () => {
```
jefBinomed: not enough tests

packages/ng/src/app/tree-picker/standalone-tree/tree.component.ts
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	/** Returns all the nodes which are checked in the given tree */
+	private getCheckedNodes(tree: Tree): ITreeNode[] {
+		const result = new Array<ITreeNode>();
+		tree.traverse(current => {
+			if (current.node.isSelected) {
+				result.push(current.node);
+			}
+			return true;
```
jefBinomed: as you return true, you will be stop at the first selected node ? so you won't get all the check nodes ?

packages/ng/src/app/tree-picker/standalone-tree/tree.message.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Subject } from 'rxjs/Subject';
+import { Observable } from 'rxjs/Observable';
+import { TreeNode } from './tree.class';
+
+export class TreePickerMessage {
```
jefBinomed: Add some Description about the rule of this class

packages/ng/src/app/tree-picker/tree.module.ts
```diff
@@ -0,0 +1,32 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { FormsModule } from '@angular/forms';
+import { LuTreeComponent } from './standalone-tree/tree.component';
+import { HttpClientModule } from '@angular/common/http';
+import { OverlayModule } from '@angular/cdk/overlay';
+import { LuTreeItemComponent } from './standalone-tree/tree-item/tree-item.component';
```
jefBinomed: the package `standalone-tree` could be replace with just `tree` ?

packages/ng/src/app/tree-picker/picker/tree-picker.component.html
```diff
@@ -0,0 +1,23 @@
+<div [LuPopoverTriggerFor]="treePickerPopover">
+	<span class="single-value" *ngIf="multiple === false && value != null">
+		{{ value.name }}
+	</span>
+	<span class="single-value-empty" *ngIf="multiple === false && value?.id == null">
+		{{ placeholder }}
```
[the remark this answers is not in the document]
bastienguillon: There's a placeholder for each different case:
+ multiple empty: just display the placeholder
+ multiple not empty: display the original placeholder + a small box with the number of selected items (see the screenshot in the PR description)
+ single empty: just display the placeholder
+ single not empty: display the name of the selected item

'multiple empty' and 'single empty' could be regrouped with only one condition, that's true

packages/ng/src/app/tree-picker/picker/tree-picker.component.ts
```diff
@@ -0,0 +1,80 @@
...
+	@Input()
+	public multiple: boolean;
+	/** The placeholder of the component, displayed when:
+	 * - No value is selected
+	 * - A value is selected and `multiple` is set to true
+	 */
+	@Input()
+	public placeholder: string;
+
+	public _innerValue: any;
```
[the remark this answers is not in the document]
bastienguillon: `_innerValue` should be `private`, indeed. I'll change its type too

packages/ng/src/app/tree-picker/picker/tree-picker.component.ts
```diff
@@ -0,0 +1,80 @@
...
+
+@Component({
+	selector: 'lu-tree-picker',
+	templateUrl: './tree-picker.component.html',
+	providers: [
+		{ provide: NG_VALUE_ACCESSOR, useExisting: forwardRef(() => LuTreePickerComponent), multi: true },
+	],
+	styleUrls: ['./tree-picker.component.scss'],
+})
+export class LuTreePickerComponent implements ControlValueAccessor, OnInit, OnChanges {
```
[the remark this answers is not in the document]
bastienguillon: The model just have to implement the `ITree` interface, generics wouldn't be useful here

packages/ng/src/app/tree-picker/picker/tree-picker.component.ts
bastienguillon: I did this to avoid having to check for `null` or `undefined` every time I manipulate the value. The `_innerValue` used in the component will not be null, however the ngModel will stay `null` or `undefined`: it has no consequence for the user of the component

packages/ng/src/app/tree-picker/standalone-tree/tree-item/tree-item.component.spec.ts
bastienguillon: I was waiting for comments on this PR

packages/ng/src/app/tree-picker/standalone-tree/tree.component.html
bastienguillon: I don't think so: the `lu-tree` component is in charge of managing the ngModel, whereas the `lu-tree-item` components only manipulate the internal `tree` attribute

LuccaIntegration:
Travis automatic deployment: https://227-pr-lucca-front-LuccaSA.surge.sh

bastienguillon:
Too many changes are required to add these components to Lucca-Front and the `select` component is not mature enough right now to make any long-term decision. I close this PR, however...

giphy

bastienguillon:
Do not delete this branch btw

PR #229, opened 2018-02-21
[SCSS][ENH] navSide - Better style after integration

sanlucca:
Style changes after testing it in prod. It now uses margin-bottom rather than padding and adds a z-index. It is necessary to do so to work on both FF and IE.


LuccaIntegration:
Travis automatic deployment: https://229-pr-lucca-front-LuccaSA.surge.sh

PR #233, opened 2018-02-22
Ng.build.prod

lucienbertin:
trimmed until `npm run ng build -prod -a=demo` worked

flagged breaking changes with :boom:

 - mini refacto of animations file names
 - :boom: renamed `LFAnimationFactory` to `LfAnimationFactory` for consistency 's sake
 - :boom: changed `LfAnimationFactory` signature to be aot compatible (see after break)
 - use factory to provide animation so it supports `aot`
 - fix selects module to avoid declaring components twice
 - :boom: couldn't keep using popover.template.ts file cuz `-aot`, as a result had to create a popover.component.html. As a result people using `customPopoverTemplate` (api-picker or select-picker) can still use it but it wont work in aot. so you have to create the html file and have the whole template in it - see [this](https://github.com/LuccaSA/lucca-front/commit/d02a2e974a485bf2b4010e2e61dbc4ac1cc05636) commit
 - fixed the demo in aot
 - rem demo/formly/debug
 - fixed aot issues in select due to `protected/private` properties being used in html template

-----------

in [this commit](https://github.com/LuccaSA/lucca-front/pull/233/commits/9aeca8de9efa0f0d15ac76eee85007aa712d3961), i changed `LfAnimationFactory` signature from 
```ts
function LfAnimationFactory(type: strng, duration: number, delay: number, easing: string): AnimationTriggerMetadata
```
to
```ts
 function LfAnimationFactory(type: strng, animationTiming: string): AnimationTriggerMetadata
```
because the animation timing was computed inside the function as 
```ts
const animationTiming = `${duration}ms ${delay}ms ${easing}`;
```
problem is that string construction via ` is not compatible AOT


LuccaIntegration:
Travis automatic deployment: https://233-pr-lucca-front-LuccaSA.surge.sh

lucienbertin:
@jefBinomed can you add a travis step to make sure the build -aot keeps on working

lucienbertin:
wait - demo/animation is broken, i'm on it

LuccaIntegration:
Travis automatic deployment: https://233-pr-lucca-front-LuccaSA.surge.sh

packages/ng/src/app/select/select.module.ts
```diff
@@ -1,9 +1,9 @@
 import { NgModule } from '@angular/core';
 import { CommonModule } from '@angular/common';
 import { OverlayModule } from '@angular/cdk/overlay';
-import { LuSelectPicker } from './picker';
-import { LuSelectOption } from './option';
-import { LuSelectClearerComponent, LuSelectClearerFirstOrDefaultComponent } from './clearer';
+import { LuSelectPickerModule } from './picker';
+import { LuSelectOptionModule } from './option';
+import { LuSelectClearerModule } from './clearer';
 import { LuSelectDirective } from './directive';
```
jefBinomed: Why the directive isn't import in the same way as the others modules ?

lucienbertin: i didnt see select/sdirective had a module, i'll fix that asap

LuccaIntegration:
Travis automatic deployment: https://233-pr-lucca-front-LuccaSA.surge.sh

PR #237, opened 2018-02-27
CI stuff (AOT, travis)

lucienbertin:
this pr contains :

 - :hammer: travis calls script `travis` from root package, it is then dispatched to lf packages via lerna (for now only ng package is tested)
 - :hammer: ng - travis script does this
   - builds the demo in aot
   - run tests for /src
   - builds the demo in JIT, no minification/cachebusting for easier reverse engineering
 - :boom: renammed `LfAnimationFactory` to `LuAnimationFactory` for consistency
 - :bug: user-tile - use the display pipe the right way, is compatible aot

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,7 @@ addons:
 before_script:
 
 script:
-  - cd packages/ng
-  - npm test
-  - gulp generatedoc
-  - ng build -a=demo
-  - cd ../../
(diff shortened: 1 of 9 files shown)


LuccaIntegration:
Travis automatic deployment: https://237-pr-lucca-front-LuccaSA.surge.sh

PR #257, opened 2018-03-29
Scss.icons default path

jeremie-lucca:


LuccaIntegration:
Travis automatic deployment: https://257-pr-lucca-front-LuccaSA.surge.sh

jeremie-lucca:
Each project on LF have an icon URL override on cdn to avoid a new font loading on each project. 
We hesitated to put icons on statics, but as LF is a public project, cdn looked to be a better solution.

PR #258, opened 2018-03-29
api feeder component

lucienbertin:
minor refactor of the /seclect/api directory to add a rddapifeeder component so i can have this way of linking the select to its feeder

```html
<lu-api-select [selectApiFeeder]="feeder">
	<lu-api-feeder #feeder [api]="api" [params]="params"></lu-api-feeder>
</lu-api-select>
```


packages/ng/src/app/select/api/select-api.component.ts
```diff
@@ -29,7 +30,7 @@ import { ISelectApiFeeder } from './picker';
 	],
 })
 // tslint:disable-next-line:component-class-suffix
-export class LuSelectApi<T>
+export class LuSelectApi<T extends IApiItem = IApiItem>
```
jefBinomed: I'm not agree with this, because, the LuSelectApi should as generic as possible, By forcing a cast in IApiItem, you force a kind of api. So for RDDFeeder, it makes sens but not for generic LuSelect API, can you please remove this `extends`

PR #270, opened 2018-04-12
[fix] line borders gap

jeremie-lucca:
Fix this issue : 
image

### DIFF
--- a/packages/scss/src/components/_table.scss
+++ b/packages/scss/src/components/_table.scss
@@ -117,7 +117,7 @@
 		transition: opacity _theme("commons.animations.durations.fast") ease;
 		white-space: nowrap;
 
-		* {
+		> * {
 			opacity: 0;
 		}
 	}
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
Travis automatic deployment: https://270-pr-lucca-front-LuccaSA.surge.sh

packages/scss/src/components/_table.scss
```diff
@@ -110,22 +109,27 @@
...
-		opacity: 0;
 		padding: {
 			bottom: .3rem;
 			top: .3rem;
 		}
 		text-align: right;
 		transition: opacity _theme("commons.animations.durations.fast") ease;
 		white-space: nowrap;
+
+		* {
```
sanlucca: I guess this is to have better control on what is invisible ? If so, shouldn't you use a ` > *` or a specific class to only affect direct children ? It feels like it could change opacity for a lot for things this way

jeremie-lucca: Without border collapse TD's border was also invisible.
I can also use a direct selector. But as we are inside a very specific class, mod-action (which format line actions), using another class seems to be a bit painful to use.

LuccaIntegration:
Travis automatic deployment: https://270-pr-lucca-front-LuccaSA.surge.sh

PR #279, opened 2018-04-30
[SCSS] CSSComb

sanlucca:
Rewriting the "format-scss" of the root package. Il uses CSSComb and matches the stylelint of the SCSS package.


sanlucca:
Replaced by #280

PR #302, opened 2018-05-15
Migrate.demo

GuillaumeNury:


packages/ng/demo/app/shared/api-docs/api-docs-config.component.ts
```diff
@@ -20,10 +20,13 @@ export class DemoApiDocsConfig {
 	apiDocs: ClassDesc;
 	directiveName: string;
 
-	constructor() {}
+	constructor() {
+		console.log(docs);
```
lucienbertin: 😱

packages/ng/demo/app/shared/api-docs/api-docs-config.component.ts
```diff
@@ -20,10 +20,13 @@ export class DemoApiDocsConfig {
...
+	}
 
 	@Input()
 	set type(typeName: string) {
+		console.log(docs);
```
lucienbertin: :scream: :scream: :scream:

lucienbertin: its on the old code

PR #303, opened 2018-05-15
[fix] still has all border with browser zoom

jeremie-lucca:
@sanlucca Textfields looked better for me with a height of 2px less, so I did not edited component's padding.


LuccaIntegration:
Travis automatic deployment: https://303-pr-lucca-front-LuccaSA.surge.sh

PR #311, opened 2018-05-16
Formly.version

lucienbertin:
follow up of PR #308 and #310 

-------------

 - updated formly version
 - updated package.json for entry points /formly and /material

> :warning: some shenanigans with scss build 
> for some reason that i dont wanna look for, when building with ng-packagr, it seems node-sass cant resolve `@import '~@lucca-front/scss/src/theming';`
> cuz `theming` can be the flder `/theming` or the file `_theming`
> so i have to specify `@import '~@lucca-front/scss/src/theming.scss';` (n /libs/xxx/src/style/_theming.scss) or the libs tont build
> **BUT**
> when i build the demo and import `/libs/xxx/src/style/main` via the /demo/style.scss, it does'nt build if i specify the .scss
> as a result me and sandy we gonna rename the folder `/theming` -> `/theme` and hope it fixes [everything](https://media3.giphy.com/media/wLXo0vTZSM7GU/giphy.gif)

PR #314, opened 2018-05-17
Fix.build.ng

lucienbertin:
clean scss import to scss/theming that were problematic (see pr #311 )


lucienbertin:
also fixes #290

lucienbertin:
for some reason the **26th** u-test fails, it does not matter which test it is, the 26th fails

PR #313, opened 2018-05-17
[new] Add support for CSS var generation and usage based on current t…

lucienbertin:
…heming

PR #315, opened 2018-05-17
Rc.1.0

lucienbertin:
better package distribution now that ng-cli supports libraries

contains:

 - [x] migration to angular 6
 - [x] distribution as angular lib (with peer dependency and all that
 - [x] extracted formly and material dependencies from /ng package and added it as secondary entry point
 - [x] #313 theming as css vars
 - [x] clean up code in /ng folder
 - [ ] update changelog
 - [x] #316 adapt lerna scripts to publish the content of `packages/ng/dist` and not just `/ng`
 - [ ] fix travis integration

----

before merging, i'll try publishing a 1.0.0-rc.0 and migrate a project to this version, to ensure we're not publishing a broken distrib


LuccaIntegration:
Travis automatic deployment: https://315-pr-lucca-front-LuccaSA.surge.sh

PR #316, opened 2018-05-17
1.0.lerna

lucienbertin:
updated lerna.json and other stuff so lerna can discover `@lucca-front/ng` which is under `/packages/ng/dist` and dont publish `ng-internal` which used to be `/packages/ng`

### DIFF
--- a/packages/ng/package.json
+++ b/packages/ng/package.json
@@ -6,6 +6,7 @@
     "type": "git",
     "url": "git+https://github.com/LuccaSA/lucca-front.git"
   },
+  "private": true,
   "keywords": [
     "lucca",
     "angular",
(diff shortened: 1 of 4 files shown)


packages/ng/package.json
```diff
@@ -6,6 +6,7 @@
     "type": "git",
     "url": "git+https://github.com/LuccaSA/lucca-front.git"
   },
+  "private": true,
```
jefBinomed: Why Private ? Lucca-Front isn't supposed to be public ?

lucienbertin: its not this package.json that gets published, this one serves as an entry point to build /librairies, then we publish whats inside /dist

lucienbertin: run `yarn build` then look whats inside `/dist`

lucienbertin: @jefBinomed ??

jefBinomed: nothing more ;) I didn't reopen the PR during the whole day

PR #333, opened 2018-07-02
aadd default format for user display pipe

lucienbertin:

PR #336, opened 2018-07-16
Ng.usertile.defaultformat

lucienbertin:
same as #333 but for the user tile

it sets the display format as undefined so it falls back on `userDisplay`'s default format


LuccaIntegration:
Travis automatic deployment: https://336-pr-lucca-front-LuccaSA.surge.sh

PR #346, opened 2018-07-27
Scroll.directive

lucienbertin:
adds a scroll directive and uses it in the option picker to ping all operators


LuccaIntegration:
Travis automatic deployment: https://346-pr-lucca-front-LuccaSA.surge.sh

PR #347, opened 2018-07-30
Picker.events

lucienbertin:
adds way for the picker to tell the operators some events are occuring
like keyboard evvents, picker opening/closing


LuccaIntegration:
Travis automatic deployment: https://347-pr-lucca-front-LuccaSA.surge.sh

PR #349, opened 2018-08-02
Select.display

lucienbertin:
add a displayer interface with 2 methods
- getELtRef(value)
- getViewRef(value)

then the selects tries displaying the value as view first, then as element ref

displaying with the elementref uses the renderer and nativeelt so its not super duper great


LuccaIntegration:
Travis automatic deployment: https://349-pr-lucca-front-LuccaSA.surge.sh

LuccaIntegration:
Travis automatic deployment: https://349-pr-lucca-front-LuccaSA.surge.sh

PR #353, opened 2018-08-03
Api.select

lucienbertin:
adds component `<lu-api-select>` which is a generic version of what you can do with the select + option picker + api-operator


LuccaIntegration:
Travis automatic deployment: https://353-pr-lucca-front-LuccaSA.surge.sh

PR #356, opened 2018-08-06
fix demo

lucienbertin:
fixes the demo for /api/select and /user/select


LuccaIntegration:
Travis automatic deployment: https://356-pr-lucca-front-LuccaSA.surge.sh

PR #361, opened 2018-08-07
Input.directive

lucienbertin:
add 
- ILuInput<T> - pretty empty interface that extends `ControlValueAccessor`
- ALuInput<T> - adds class `is-filled` when value is not empty
- LuInputDirective<T> - extends ALuInput
- ALuSelect - implements ILuInput<T> and add `is-filled` class to host element - affects all selects (normal, api, user, ...)


LuccaIntegration:
Travis automatic deployment: https://361-pr-lucca-front-LuccaSA.surge.sh

PR #370, opened 2018-08-21
option-picker - add class `is-selected` to current value

lucienbertin:
when the value chages for the select-input (control value accerssor) it passes the info to its picker so the picker adds the class `is-selected` to current value


LuccaIntegration:
Travis automatic deployment: https://370-pr-lucca-front-LuccaSA.surge.sh

Issue #375, opened 2018-08-27
luinput directive with another CVA


lucienbertin:
you cant have 2 CVA on same component, so luinput doesnt work on tag input, which is a bit ironic

should write it as a control value _reader_ not accessor. like the directive [luEmpty](https://github.com/LuccaSA/lucca-front/blob/master/packages/ng/libraries/core/src/lib/empty/empty.directive.ts)

lucienbertin:
fixed in #379

PR #379, opened 2018-08-31
Input.directive

lucienbertin:
when there are 2 control value accessor on the same tag, it throws an error and thats kind of normal

so luInput isnt a cva anymore, cuz it just reads the value, it doesnt write it


LuccaIntegration:
Travis automatic deployment: https://379-pr-lucca-front-LuccaSA.surge.sh

lucienbertin:
fixes #375

PR #380, opened 2018-08-31
Api.select.fix

lucienbertin:
- fix the bug i showed you with the picker being all white
- fix the class `is-selected` not being reapplied when the option displed changed


LuccaIntegration:
Travis automatic deployment: https://380-pr-lucca-front-LuccaSA.surge.sh

PR #393, opened 2018-09-27
Fix.lu select

### LINKED ISSUE #392: lu-select do not display 0
it seems that falsy values are not passed to the displayer

GuillaumeNury:
Fix #392 	
Fix cannot read property element of undefined


LuccaIntegration:
Travis automatic deployment: https://393-pr-lucca-front-LuccaSA.surge.sh

PR #396, opened 2018-10-02
Scss.animation.theming

jeremie-lucca:


LuccaIntegration:
Travis automatic deployment: https://396-pr-lucca-front-LuccaSA.surge.sh

packages/scss/src/utilities/animations/_sliding.scss
```diff
@@ -1,7 +1,7 @@
 @keyframes slideIn {
```
lucienbertin: slideIn seems to be a shortcut for slideInTop. is it on purpose ?

lucienbertin: by that i mean we got
- slideIn
- slideInLeft
- slideInRight
- slideInBottom

but no slideInTop

sanlucca: Yes, as Top was seen as default

PR #431, opened 2018-10-08
[ng][fix] Select - isEmpty if empty array

sanlucca:
Fix the issue of empty array of values not being considered as an empty select


LuccaIntegration:
Travis automatic deployment: https://431-pr-lucca-front-LuccaSA.surge.sh

PR #437, opened 2018-10-11
[NG][FIX] Select disabled state

sanlucca:
Selects should use `disabled` attribute rather than `is-disabled` class


LuccaIntegration:
Travis automatic deployment: https://437-pr-lucca-front-LuccaSA.surge.sh

PR #450, opened 2018-10-31
[feat] rxjs pipe synthax

### LINKED ISSUE #332: rxjs - use the pipe syntax
we're still using the old synthax of rxjs, that means that we need to depend on `rxjs-compat`

we should migrate to the `pipe`/`tap` syntax instead

it should be invisible for the end user

needed for angular 7 cuz the old synthax is not supported by rxjs 6.3 onwards

dont forget to update peer dependencies in package.json ([here](https://github.com/LuccaSA/lucca-front/blob/master/packages/ng/libraries/core/package.json#L9), [here](https://github.com/LuccaSA/lucca-front/blob/master/packages/ng/libraries/formly/package.json#L10) and [here](https://github.com/LuccaSA/lucca-front/blob/master/packages/ng/libraries/material/package.json#L8))

lucienbertin:
removed old rxjs synthax
```ts
import 'rxjs/add/operator/map';
obs$.map(() => {});
```
in favor of the pipe synthax
```ts
import { map } from 'rxjs/operators';
obs.pipe(map(() => {}));
```
and removed dependency and peer dependency to `rxjs-compat`

closes #332


LuccaIntegration:
Travis automatic deployment: https://450-pr-lucca-front-LuccaSA.surge.sh

LuccaIntegration:
Travis automatic deployment: https://450-pr-lucca-front-LuccaSA.surge.sh

LuccaIntegration:
Travis automatic deployment: https://450-pr-lucca-front-LuccaSA.surge.sh

PR #462, opened 2018-11-15
[fix] lu-select clearer causing unexpected form submit

### LINKED ISSUE #461: Select clearer causes form submit
Select clearer template is a button without type. Since button default type is submit, if select with clearer is used inside a form tag, clicking on the clearer causes an unexpected form submit.

bmoriniere:
fixes #461 

lu-select clearer was causing unexpected form submit, because button default type is submit.


LuccaIntegration:
Travis automatic deployment: https://462-pr-lucca-front-LuccaSA.surge.sh

PR #465, opened 2018-11-19
Fix.select.disabled

### LINKED ISSUE #457: lu-select disabled input unexpected behavior
When using lu-select with a disabled input bind to variable, lu-select is not disabled as expected.

Let's say we have a disabled var :
- If binding is `[attr.disabled]="disabled ? true : null"` : If reflects well changes of the variable, except on initial value : if initial value of disabled is true, select is not disabled.
- If variable is binded to`[disabled]="disabled"` or `[disabled]="disabled ? true : null"` : 
if initial value is true and then becomes false, lu-select stays disabled.
if initial value is false and we switch value multiple times, at some point it becomes disabled and never comes back.
- finally  `[attr.disabled]="disabled"` is always disabled as expected.

Reproduced  in demo, in foundations > basic : add a disabled variable to component, in template add somewhere `<input type="checkbox" [(ngModel)]="disabled">`, and then test different disabled bindings on lu-select

lucienbertin:
hotfix for you bug boris

closes #457 

there are still some shenanigans happening

### DIFF
--- a/packages/ng/libraries/core/src/lib/select/input/select-input.component.ts
+++ b/packages/ng/libraries/core/src/lib/select/input/select-input.component.ts
@@ -78,7 +78,7 @@ implements ControlValueAccessor, ILuInputWithPicker<T>, AfterContentInit {
 			_renderer,
 		);
 	}
-	@HostBinding('attr.disabled')
+	@HostBinding('class.is-disabled')
 	get isDisabled() { return this.disabled; }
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
Travis automatic deployment: https://465-pr-lucca-front-LuccaSA.surge.sh

jeremie-lucca:
:shipit:

PR #467, opened 2018-11-23
ng-formly - fix - api-select and user seklect diodnt support multiple

lucienbertin:
formly fields of type `api` and `user` didnt support template-option _multiple_


LuccaIntegration:
Travis automatic deployment: https://467-pr-lucca-front-LuccaSA.surge.sh

PR #471, opened 2018-11-28
[fix] ie 11 compatibility

lucienbertin:
the problem was that ie 11 didnt support css vars,
so we used a polyfill,
but due to angular injecting things with shadow dom or something, newly injected at runtime comnponent didnt get the polyfill treatment
so it didnt work super well


LuccaIntegration:
Travis automatic deployment: https://471-pr-lucca-front-LuccaSA.surge.sh

PR #515, opened 2018-12-07
[feat] ng-formly - api allow orderby

### LINKED ISSUE #509: ng-formly - api - allow change of order by in the template options
it would be nice to be able to change the default order by when using api-select via formly because

- not all apis support orderby=name,asc (and it makes me sad)
- for some things, a different order by is more pertinent (no i dont have an example)

Oxthan:
closes #509


LuccaIntegration:
Travis automatic deployment: https://515-pr-lucca-front-LuccaSA.surge.sh

PR #516, opened 2018-12-07
[fix][SCSS] NavSide moving when scrolling on IE11

sanlucca:
Fixes the issue where the navSide with a `navSide-bottomSection` was jumping when scrolling on IE.
We introduce a `navSide-mainSection` that wraps around the `navSide-scrollWrapper`.


LuccaIntegration:
Travis automatic deployment: https://516-pr-lucca-front-LuccaSA.surge.sh

PR #526, opened 2018-12-13
[FIX][SCSS] Mobile menu height should be toggle size.

sanlucca:
Padding-top on mobile menu must be toggle size and not banner size.


LuccaIntegration:
Travis automatic deployment: https://526-pr-lucca-front-LuccaSA.surge.sh

PR #531, opened 2018-12-18
[fix] [SCSS] Checkboxesfield mod-row was not properly defined

sanlucca:
Checkboxesfield mod-row used `.radio` instead of `.checkbox`
Both checkboxesfield & radiosfield now uses a common variable for horizontal-spacing


LuccaIntegration:
Travis automatic deployment: https://531-pr-lucca-front-LuccaSA.surge.sh

PR #545, opened 2019-01-10
Fix.scss2.2

jeremie-lucca:


packages/ng/libraries/material/src/style/components/_datepicker.scss
```diff
@@ -6,42 +6,47 @@
 }
 
 .mat-datepicker-toggle {
-	@extend .textfield-suffix;
```
sanlucca: Why removing the extend and duplicating the style ?

packages/scss/src/components/_action-icon.scss
```diff
@@ -27,10 +27,9 @@
 
 .actionIcon {
 	&.mod-invert {
-		color: #CCC;
+		color: white;
```
sanlucca: You might want to make it a variable. If I'm correct, this is used to display actionIcon on dark bg. Not sure white will always be our go-to pick. What do you think ?

packages/ng/libraries/material/src/style/components/_datepicker.scss
jeremie-lucca: Because we added a pointer-event: none for suffixes (a click on search icon inside a textfield prevented the focus ), but datepicker still need to be clickable.

packages/scss/src/components/_action-icon.scss
jeremie-lucca: Yup, I hesitated...
Let's add it on theming.

packages/ng/libraries/material/src/style/components/_datepicker.scss
sanlucca: then you might want to still use the extend and then change the pointer-events afterwards. This allows to keep a common base of styling

LuccaIntegration:
Travis automatic deployment: https://545-pr-lucca-front-LuccaSA.surge.sh

LuccaIntegration:
Travis automatic deployment: https://545-pr-lucca-front-LuccaSA.surge.sh

PR #569, opened 2019-02-19
[fix] Container width wasn't properly calculated

sanlucca:
Container responsive width was broken as it wasn't calculated but juste putting value in a string


LuccaIntegration:
Travis automatic deployment: https://569-pr-lucca-front-LuccaSA.surge.sh

PR #578, opened 2019-02-28
picker rescroll shenanigans

### LINKED ISSUE #577: option-picker - rescroll at opening
when opening an option picker with a value in the model, it tries to scroll to said value. when using an api pager or paged searcher it tries to scroll to bottom of the page, that triggers a call for next page of results, and if the picker doesnt find the value in the 40 available results, it triggers another call but with page 0 this time

it shouldnt try to scroll to bottom if the selected value isnt in the first page

how to reproduce
- go on the demo
- open select using an api pager or paged searcher
- select a value in page 3
- reopen picker
- see that 3 calls have spawned, for paage 0, then page 1, then page 0 again

lucienbertin:
fixes #577 

-----

after some debug it seemed that the rescroll was triggered by the fact that at the time of the opening of the picker, old options were still there, so it could find the highlighted option and it tried to scroll to it. it was due to option list reinitialization was made on the onOpen event, add some debouncetime and you have this

- i have 80 options and select the 77th
- i close the picker
- i reopen the picker, all 80 options are available
- it highlights the 77th, and try to scroll to it
- at the same time onOpen is called in the api-pager
- it reinits with just 20 options
- it tries to scroll to the 77th of the 20 options
- all hell breaks loose

so the fix was to reinit the option list on close

### DIFF
--- a/packages/ng/libraries/core/src/lib/api/select/pager/api-pager.model.ts
+++ b/packages/ng/libraries/core/src/lib/api/select/pager/api-pager.model.ts
@@ -2,7 +2,7 @@ import { ILuOptionOperator } from '../../../option/index';
 import { IApiItem } from '../../api.model';
 import { HttpClient } from '@angular/common/http';
 import { Observable, of, merge, Subject } from 'rxjs';
-import { switchMap, catchError, mapTo, tap, debounceTime, map, distinctUntilChanged } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
Travis automatic deployment: https://578-pr-lucca-front-LuccaSA.surge.sh

PR #601, opened 2019-04-02
Fix.options.width

jeremie-lucca:
Allow option select to be wider than select.
(With help of @GuillaumeNury <3)

image


LuccaIntegration:
Travis automatic deployment: https://601-pr-lucca-front-LuccaSA.surge.sh

PR #613, opened 2019-04-10
[bug][ng] Adds a loader when opening an option-picker

### LINKED ISSUE #609: [api select] Dropdown can be offscreen
Our dropdown has a bigger height after loading. If the loader fits under the select, the dropdown will stay under, even if its height doesn't fit on the viewport.

sanlucca:
Fixes #609 
It adds a loader when opening an option-picker that takes 100% of the picker max-height. Therefor, the initial height of the select is more accurate and the dropdown is correctly placed by the overlay service.

### DIFF
--- a/packages/ng/libraries/core/src/lib/option/picker/option-picker.model.ts
+++ b/packages/ng/libraries/core/src/lib/option/picker/option-picker.model.ts
@@ -1,6 +1,6 @@
 import { ILuPickerPanel, ALuPickerPanel } from '../../input/index';
 import { Subscription, Observable, merge } from 'rxjs';
-import { switchMap } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


LuccaIntegration:
Travis automatic deployment: https://613-pr-lucca-front-LuccaSA.surge.sh

packages/ng/libraries/core/src/lib/option/picker/option-picker.model.ts
```diff
@@ -11,6 +11,8 @@ export abstract class ALuOptionPicker<T = any> extends ALuPickerPanel<T> impleme
 	private __operators;
 	protected _subs = new Subscription();
 	onSelectValue: Observable<T | T[]>;
+	loading$: Observable<boolean>;
+	isEmpty$: Observable<boolean>;
```
lucienbertin: you dont seem to need this one

PR #615, opened 2019-04-10
[bug][ng] Select padding and picker minwidth

### LINKED ISSUE #610: [bug] lu-api-select and lu-user-select should always have a bigger right padding
`lu-api-select` and `lu-user-select` are always clearable. Therefore it should have a bigger right padding as the is-clearable state for textfield-input

sanlucca:
Fixes #610
`api-select` and `user-select` are always clearable. Therefor, we are adding a similar padding as on the `is-clearable` state .
#601 broke the fact that the picker should at least be as wide as its input. It is now the case, yet we force it to be at least 185px wide.


LuccaIntegration:
Travis automatic deployment: https://615-pr-lucca-front-LuccaSA.surge.sh

PR #618, opened 2019-04-16
[fix] option-picker async wasnt focusing search input on open

lucienbertin:
with the addition of the loader in patch 2.5.1, when the loading was done the search field of an option picker with a search operator wasnt focused

to fix it we replay the onOpen event at the end of loading


LuccaIntegration:
Travis automatic deployment: https://618-pr-lucca-front-LuccaSA.surge.sh

PR #656, opened 2019-06-04
[NG][SCSS] Fix textarea disabled on mod-framed

sanlucca:
Textarea disabled wasn't working on mod-framed because in this mod, the `textfield` needs to be the one disabled, not the input. 
Therefore, I adapted the style in the `scss` package and added a `is-disabled` on the `textfield-layout` for formly


LuccaIntegration:
Travis automatic deployment: https://656-pr-lucca-front-LuccaSA.surge.sh

PR #668, opened 2019-06-17
[fix] adds state class for is disabled is readonly

sanlucca:
Fixes disabled state being overriden for material and framed style


LuccaIntegration:
Travis automatic deployment: https://668-pr-lucca-front-LuccaSA.surge.sh

PR #672, opened 2019-06-18
Refacto.option item.selected

lucienbertin:
# BREAKING
added 
```ts
	selected: boolean;
	highlighted: boolean;
```
in the ILuOptionItem interface/abstract class

----

added ngIf in tree option item icons for selecting parent/children only

changed the way the picker flags the selected/highlighted option(s). 
before it used the renderer to add class `is-selected` to each selectd option and remove it from unselected ones.
now it set `option.selected = true/false` and its the responsibility of the option to apply its style according to its status

the advantages of this approach are
- the part with the renderer was a bit :puke:, thats good that thats gone
- for complex option-item component, you dont necessarily want the class `is-selected` on the host DOM element, here you can apply it however you want [*]

> [*] for example the tree-option-item want to apply the class only to the node displaying the parent, not the one with the children too


LuccaIntegration:
Travis automatic deployment: https://672-pr-lucca-front-LuccaSA.surge.sh

LuccaIntegration:
Travis automatic deployment: https://672-pr-lucca-front-LuccaSA.surge.sh

PR #673, opened 2019-06-19
[add] Scroll block and scroll in sidepanel

sanlucca:
Adds sizing for modals/popup
Adds scrollblock from cdk
Changes scroll strategy in the sidepanel where the header scrolls with the content. It avoids craming the scroll between to heavy blocks.


LuccaIntegration:
Travis automatic deployment: https://673-pr-lucca-front-LuccaSA.surge.sh

PR #676, opened 2019-06-21
Rem node sass

### LINKED ISSUE #652: Make @lucca-front/scss compatible with sass package
Until Angular 7, [node-sass package](https://www.npmjs.com/package/node-sass) is the default sass builder for Angular apps. Since Angular 8, [sass package](https://www.npmjs.com/package/sass) is the default one.

There are some differences between sass and node-sass listed [here](https://www.npmjs.com/package/sass#behavioral-differences-from-ruby-sass).

lucienbertin:
i did this branch where i removed everything related to node-sass

- package scss now builds with `gulp-dart-sass` that uses the npm package `sass`
- package ng - removed dependency and peer dep to node-sass

so now when you use `npm start` in packasge `scss`, it works the same but uses `sass` and not `node-sass`

also when you use `npm start` in package `ng`, it builds with `sass`. as a result it doesn't build cuz of some `Compound selector` in `@lucca-front/ng/material` stylesheet. but once all bugs are ironed out it will build

@sanlucca i let you finish this while i'm on vacation

------

fixes #652

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -860,11 +860,11 @@
       "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw=="
     },
     "@octokit/endpoint": {
-      "version": "5.1.7",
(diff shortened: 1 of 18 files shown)


LuccaIntegration:
Travis automatic deployment: https://676-pr-lucca-front-LuccaSA.surge.sh

PR #688, opened 2019-07-15
[fix] Select picker overlap

lucienbertin:
with the refactor of the popovers, it was not possible anymore to make  the picker of a select overlap its input. now its possible again

was used in [prisme](http://prisme.lucca.io/home) for the search bar


LuccaIntegration:
Travis automatic deployment: https://688-pr-lucca-front-LuccaSA.surge.sh

PR #712, opened 2019-07-25
option items - for some reason lu-option-item had a clientHeight of 0…

lucienbertin:
…, as a result it broke the scroll to highlighted in option picker

see https://github.com/LuccaSA/ilucca/pull/11376#issuecomment-515061269


LuccaIntegration:
Travis automatic deployment: https://712-pr-lucca-front-LuccaSA.surge.sh

PR #721, opened 2019-07-30
[link] Should forced to be inline or inline block

almonais:
#693


LuccaIntegration:
Travis automatic deployment: https://721-pr-lucca-front-LuccaSA.surge.sh

PR #735, opened 2019-08-12
tree-picker add missing exports for override

lucienbertin:
add missing export for a clean override of tree picker


LuccaIntegration:
Travis automatic deployment: https://735-pr-lucca-front-LuccaSA.surge.sh

PR #745, opened 2019-08-27
[feat] modal dismiss

### LINKED ISSUE #741: [luModale] ILuModalContent.dismissAction
Sometime we need to run specific code after clicking on the "Dimiss" button in a modale (e.g. dispatch an ngrx event/command). It is not possible to do this with the current `LuModal` implementation.

Adding a `dismissAction` method to the `ILuModalContent` interfaceand call it when clicking on the "Dismiss" button or the top right cross would be a nice addition.

lucienbertin:
fixes #741 calls cancelAction if the modal content provides any before closing


LuccaIntegration:
Travis automatic deployment: https://745-pr-lucca-front-LuccaSA.surge.sh

packages/ng/libraries/core/src/lib/overlay/modal/modal-panel.component.ts
```diff
@@ -51,13 +52,26 @@ export abstract class ALuModalPanelComponent<T extends ILuModalContent = ILuModa
...
 	}
 	dismiss() {
-		this._ref.close();
+		if (!this._componentInstance.cancelAction) {
+			return this._ref.close();
+		}
+		const result$ = this._componentInstance.cancelAction();
+		if (result$ instanceof Observable) {
+			this.dismissClass$.next('is-loading');
+			result$.subscribe(
```
GuillaumeNury: If `cancelAction` is not correctly implemented, it could never complete or emit multiple events. A `.pipe(first())` can prevent this. (same for `submit` method)

lucienbertin:
after deliberation with @sanlucca we'tre either
- not doing this\*
- not doing it that way, opting for ModalRef having `onClose` and `onDismiss` outputs, and deifferentiating it that way. the goal is to never prevnt a modal from closing, cuz thats annoying

\* by this i mean issue #741

PR #785, opened 2019-09-24
formly-radio - always use radio fields

lucienbertin:
i'll add an input type `button group` or something, but when the user want radio buttons, we should display radio buttons

alos what it did is display radios if less than 3 options, and display a button group for 3 or more option and i kinda feel it should be the orther way around


LuccaIntegration:
Travis automatic deployment: https://785-pr-lucca-front-LuccaSA.surge.sh

PR #803, opened 2019-10-03
[Fix][Option-picker] add class

### LINKED ISSUE #802: [ng][bug] `option-picker` should have a class
We can't override `option-picker`  style as it doesn't expose any specific CSS class. We should add one into its DOM

RipripKipkip:
fixes #802

### DIFF
--- a/packages/ng/libraries/core/src/lib/option/picker/option-picker.component.html
+++ b/packages/ng/libraries/core/src/lib/option/picker/option-picker.component.html
@@ -1,6 +1,6 @@
-<ng-template>
+<ng-template class="option-picker">
 	<div class="lu-popover-panel" role="dialog" [ngClass]="panelClassesMap" [class.mod-multiple]="multiple"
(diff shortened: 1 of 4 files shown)


LuccaIntegration:
Travis automatic deployment: https://803-pr-lucca-front-LuccaSA.surge.sh

packages/ng/libraries/core/src/lib/option/picker/option-picker.component.html
```diff
@@ -1,6 +1,6 @@
-<ng-template>
+<ng-template class="option-picker">
 	<div class="lu-popover-panel" role="dialog" [ngClass]="panelClassesMap" [class.mod-multiple]="multiple"
```
lucienbertin: the calss should be on this div, not its parent

the tag `ng-template` is a virtual tag used by angular, it will never be present and rendererd in the actual DOM

LuccaIntegration:
Travis automatic deployment: https://803-pr-lucca-front-LuccaSA.surge.sh

PR #913, opened 2019-11-20
FIX we remove the onPush strategy of overlay components

vergerf:
Sorry, there is no tickets that was created.
But we feel that restrict the usage of overlays only to OnPush application could be annoying some time.
Happy to discuss about it.

Edit: this is my first contribution, I read contributions guidelines but I don't know the validation process yet.


LuccaIntegration:
Travis automatic deployment: https://913-pr-lucca-front-LuccaSA.surge.sh

PR #916, opened 2019-11-26
[feat] date picker reloaded

lucienbertin:
major refactor of the calendar input and date input. both use a date adapter so it can be plugged to any date model you want (native, moment, luxon, ...)

added luDate pipe that uses the same adapter

added LuNativeDateAdapter, n implementation of date adapters for native date


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-916

PR #918, opened 2019-11-26
[feat] date - min/max inputs and validation

lucienbertin:
added inputs min max to components
- luDateInput direcive
- lu-calendar component
- lu-date-picker
- lu-date-select

and the ones being control value accessor now also implement validator with min/max

also added class `is-disabled` in lu-calendar fro items not in the min/max range, for any granularity, and added class is-today and is-active in monthly view and yearly view

it needs some style @jeremie-lucca 

image


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-918

PR #908, opened 2019-11-12
Feat/tree.shakable

lucienbertin:
fixes #787 

splits the `@lf/ng` lib into a lot of small libs to allow for tree shakability.

see #895 for poc of concept

# :bomb: Breaking changes

- LuModule has disappeared, there is no root module exporting all that exists in LF/ng
- LuOverlayModule has also disappeared, use lupopupmodule or lutooltipmodule or lumodalmodule instead
- i will break everything with the api-operators
- refactored user-searcher and user-searcher service
- lu-select-clearer becomes lu-input-clearer

# Does it work ?

## experiment

build sandbox with `commonChunk: false` and consumming the distributed libs
look at the size of generated chunks for the different issues

## before

image

what we can see here that each issue chunk is about the same size, the outlliers being the 3 issues that need to import mat-datepicker

the typical size of a chunk is 1.12 MB. in it you have 400kb dedicated to actual lf code, 100kb of `@angular/common/http`, 360kb of `@angular/cdk` and 280kb of `@angular/forms`

## after

image

we see that issues do not share the same chunk size anymore, ranging from 1.08Mb to 300kb.

the important thing to see also is that `@angular/forms` is not always present anymore which explain the discrepancy. also the size of lf code in each chunks varies from 400kb to 13kb. the code from `@angular/common/http` (100kb) is only present in chunks importing from /api or /user or /department

i'd say you could gain 25-75% of chunk size by having clean imports now

## raw stats

[stats.zip](https://github.com/LuccaSA/lucca-front/files/3846675/stats.zip)

-----

todo

- [x] select-clearer becomes input clearer and moves to /input
- [x] move onXXXSubscriber from /option to /core
- [x] complete refactor of user-select
- [ ] refactor api-select and alll the operators
- [ ] merge fomly/material style with root style, or find a way to split all styles
- [x] stats to be sure its usefull

PR #925, opened 2019-12-02
undid the refactor i did on the /user subpackage

lucienbertin:
in PR #908 i started a unnecessary and unwarranted refactor of /user (especially the ALuUserSrervice and other stuff)

it broke a lot for little gain so this PR undoes it

i will still refactor it a dn /api at some point, but not in this major


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-925

PR #927, opened 2019-12-05
Feat/datepicker placeholder

lucienbertin:
- default placeholder for date-input directive, gives insight as to what is expected to be inputed and follows locale_id
- selects (all of them) removed class `is-focused` when the picker overlaps the select


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-927

PR #928, opened 2019-12-05
luPipeNumber html inject

vergerf:
add the option to *not* add html into the string returned by luPipeNumber

### DIFF
--- a/packages/ng/libraries/core/src/lib/number/number.pipe.ts
+++ b/packages/ng/libraries/core/src/lib/number/number.pipe.ts
@@ -1,5 +1,5 @@
-import { Pipe, PipeTransform, Inject, LOCALE_ID } from '@angular/core';
 import { formatNumber, getLocaleNumberSymbol, NumberSymbol } from '@angular/common';
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core';
 
 @Pipe({
 	name: 'luNumber',
@@ -7,7 +7,7 @@ import { formatNumber, getLocaleNumberSymbol, NumberSymbol } from '@angular/comm
 })
 export class LuNumberPipe implements PipeTransform {
 	constructor(@Inject(LOCALE_ID) protected locale) {}
-	transform(number: number, precision = 2) {
+	transform(number: number, precision = 2, injectHtml = true) {
 		const formatted = formatNumber(number, this.locale, `1.${precision}-${precision}`);
 		const separator = getLocaleNumberSymbol(this.locale, NumberSymbol.Decimal);
 		const split = formatted.split(separator);
@@ -15,7 +15,11 @@ export class LuNumberPipe implements PipeTransform {
 		const decimal = split[1];
 		const hideDecimal = Math.round(number) === number;
 		if (precision > 0) {
-			return `${integral}<span class="decimal-part${hideDecimal ? ' u-hidden' : ''}">${separator}${decimal}</span>`;
+			return injectHtml
+				? `${integral}<span class="decimal-part${
+						hideDecimal ? ' u-hidden' : ''
+				  }">${separator}${decimal}</span>`
+				: formatted;
 		} else {
 			return integral;
 		}


LuccaIntegration:
Travis automatic deployment: https://928-pr-lucca-front-LuccaSA.surge.sh

packages/ng/libraries/core/src/lib/number/number.pipe.ts
```diff
@@ -1,21 +1,25 @@
...
 		const integral = split[0];
 		const decimal = split[1];
 		const hideDecimal = Math.round(number) === number;
 		if (precision > 0) {
-			return `${integral}<span class="decimal-part${hideDecimal ? ' u-hidden' : ''}">${separator}${decimal}</span>`;
+			return injectHtml
+				? `${integral}<span class="decimal-part${
+						hideDecimal ? ' u-hidden' : ''
+				  }">${separator}${decimal}</span>`
+				: formatted;
```
sanlucca: If you just return `formatted`, why do you use LuNumberPipe and not directly the angular number Pipe ?

vergerf: I've been doing that waitting for another solution.
But to use the pipe number correctly, I need to inject LOCALE_ID to the component.
I think it's not the component responsability to deal with locales (and precision).
Furthermore, LuNumberPipe already do it well :) !
Creating two different pipes would be another option (however it could be a little bit overkill)

lucienbertin: how are you using the [number pipe](https://angular.io/api/common/DecimalPipe) and luNumber pipe exactly ?
because if you are using it in `{{}}` in your html you don't have to specify the locale, it uses `locale_id` by default. the luNumber pipe also uses locale_id but it's not overridable so it makes me think that your locale_id is correctly set.

as to your statement

>I think it's not the component responsibility to deal with locales (and precision).

i might agree on the locale part (and thankfully that argument is optional), but the precision part is highly debatable

vergerf: I injected luNumber in my component and I would like to use the transform method.
[formatNumber](https://angular.io/api/common/formatNumber) do the trick but `locale: string` is a required parameter.
there is a default precision in luNumber (2) that I think is great. 
It's less code and even if I want another precision, the luNumberPipe is simpler than the angular string.

lucienbertin: i feel like you're trying to drive nails with a screwdriver, and proposing to add weight in the handle so it can drive nails better.

[angular pipes](https://angular.io/guide/pipes) exist as "a way to write display-value transformations that you can declare in your HTML". they should be crafted for this purpose.

the purpose of the luNumberPipe is to display numbers aligned on their integer part while hiding null decimal parts, it exists because nothing in vanilla angular does it, and for it to work it needs to inject html.

image

you can find its documentation on [prisme](http://prisme.lucca.io/components/misc/number)

the purpose of the luNumberPipe is not to help you format numbers in a component without thinking of the locale, if that is what you are after i recommend you create an injectable formatter that can abstract locale_id and precision for your components

```ts
@Injectable() class MyFormatter {
  constructor(@Inject(LOCALE_ID) private _locale) {}
  format(num: number, precision = 2) {
    return formatNumber(num, this._locale, `1.${precision}-${precision}`);
  }
}
@Component({/* ... */}) class MyComponent {
  constructor(private _formatter: MyFormatter) {}
  ngOnInit() {
    const text = this._formatter.format(1.2345);
  }
}
```

lucienbertin: i feel like you're trying to drive nails with a screwdriver, and proposing to add weight in the handle so it can drive nails better.

[angular pipes](https://angular.io/guide/pipes) exist as "a way to write display-value transformations that you can declare in your HTML". they should be crafted for this purpose.

the purpose of the luNumberPipe is to display numbers aligned on their integer part while hiding null decimal parts, it exists because nothing in vanilla angular does it, and for it to work it needs to inject html.

image

you can find its documentation on [prisme](http://prisme.lucca.io/components/misc/number)

the purpose of the luNumberPipe is not to help you format numbers in a component without thinking of the locale, if that is what you are after i recommend you create an injectable formatter that can abstract locale_id and precision for your components

```ts
@Injectable() class MyFormatter {
  constructor(@Inject(LOCALE_ID) private _locale) {}
  format(num: number, precision = 2) {
    return formatNumber(num, this._locale, `1.${precision}-${precision}`);
  }
}
@Component({/* ... */}) class MyComponent {
  constructor(private _formatter: MyFormatter) {}
  ngOnInit() {
    const text = this._formatter.format(1.2345);
  }
}
```

vergerf: As I said
> Creating two different pipes would be another option (however it could be a little bit overkill)
I'm sorry that I use the pipe word so I wasn't understood. 
I meant use an injectable (exactly the one you wrote above) that LuPipeNumber would rely on too.

I heard that luccaFront is aiming to have a translation system that will will spare us to do boilerplate.
I didn't what to add the `MyFormatter` myself (I went for the fast way).

>i feel like you're trying to hammer nails with a screwdriver, and proposing to add weight in the handle so it can hammer nails better.

@lucienbertin, it's the second time in two weeks that you use unpleasant analogies and I'm started to get tired of this situation. 
You said some months ago that only few people contributed to this repo (same as the old lucca-ui one), but you make it hard to do so.
(Having to debate here in english can be a first wall for instance).

PR #932, opened 2019-12-16
Poc/popover.responsive

lucienbertin:
PoC of a popover that opens as a popup when open in a mobile (or in a screen smaller that 576px here)

it should be easy to malke it work with any type of popover (select picker and whatnot) or everride it in some cases (tooltips for example shouldnt open in fullsize)

i still need some guidelines as to what comportment to change


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-932

PR #947, opened 2020-01-16
[fix] build broken after angular@9.0.0-rc.8

lucienbertin:
package ng's sandbox wasn't building after moving to ng ^9.0.0-rc.8

after diagnostic it seems that removing the complex generics in the lu-select copmponent fixes it

```ts
// doesnt compile
@Component({ /* ... */})
export class LuSelectInputComponent<T = any, P extends ILuPickerPanel<T> = ILuPickerPanel<T>> 
extends ALuSelectInput<T, P>
{ /* ... */ }

// but this does
@Component({ /* ... */})
export class LuSelectComponent<T = any> { /* ... */ }
```

the overgenerification of the select-input was not strictly speaking necessary, so for now it is removed so the library can be built and used

------

what broke it was that, i think - https://github.com/angular/angular/pull/34021


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-947

PR #953, opened 2020-02-05
[fix] select touched

lucienbertin:
- select wasnt flagged as touched when closing the picker via click outside
- changed contentquery to find the clearer to static false to allow for adding a ngif on it


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-953

PR #960, opened 2020-02-19
[add] optionComparer input on option-pickers

GuillaumeNury:
To detect which option are selected, there is a comparison of JSON values. When using API/User Select, it is more safe to compare options by id.

This PR add a new input to pass something like `(o1, o2) => o1.id === o2.id`.

If we want to go further, feeders could provide a comparer that would be used by picker.
API/Users feeder could provide a ByIdOptionComparer. When no comparer are provided, we can stick to the JSON comparer.


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-960

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-960

PR #961, opened 2020-02-19
Fix/utc.date.picker

lucienbertin:
a lot of bugs in the date-select due to the fact that the `lu-calendar` was using both js dates and the dateadapter for its operations.

also added a utc native date adapter to bind your datepicker to utc dates and not local dates


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-961

PR #966, opened 2020-03-05
[fix] select - custom displayer

lucienbertin:
depending on if ivy is enabled or not, the order in which `@contenChild` and `@ViewChild` query are executed changes and it is source of bug because i over relied on `get/set` (fyi - it was the source of the bugs that caused me to panicked when i tried to migrate to ng 9 + ivy)

now i rely on _official_ life-cycle hooks to initialise my components

----

[ici](http://lucca-front.lucca.local/rc/ng/#/split-select) on peut voir que le custom displayer est ignore alors que [la](http://lucca-front.lucca.local/PR-966/ng/#/split-select) il remarche


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-966

PR #978, opened 2020-04-08
[fix]rem  unneedeed import

lucienbertin:
it broke transpil without the right includepath


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-978

PR #1014, opened 2020-05-27
[feat] user-select me option

### LINKED ISSUE #984: [NG][User-Select] Put the active user as the first option
image

It would be added by default in the user-select.
It should not appear if you're not in the selected population

### LINKED ISSUE #1000: [NG][User-select] Allow to sort "Me" as the first option of a user-select

lucienbertin:
fixes #984 
fixes #1000

add `Me: my name` as first option in user-select
- this option is always visible, it wont disappear when you type in the search bar
- this option is only displayed if the user might appear in the rest of the select
- only the user-select is concerned

# Breaking

i did a mini rename of all interfaces interfacing with the api because they were not properly prefixed (IApiItem -> ILuApiItem ansd so on)


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1014

packages/ng/libraries/user/src/lib/select/searcher/user-searcher.model.ts
```diff
@@ -31,11 +31,17 @@ implements ILuApiPagedSearcherService<U> {
...
 	protected _get(url) {
-		return (<any>super._get(url) as Observable<ISuggestion<U>[]>)
+		return (<any>super._get(url) as Observable<ILuApiSuggestion<U>[]>)
 		.pipe(map(suggestions => suggestions.map(s => s.item)));
 	}
 	protected _clueFilter(clue) {
 		const urlSafeClue = clue.split(' ').map(c => encodeURIComponent(c)).join(',');
 		return `clue=${urlSafeClue}`;
 	}
+	getMe(): Observable<ILuUser> {
```
GuillaumeNury, line 41: Why don't we just pass a new input to user-select ? Most of the time (if not all the time), apps know who is connected. That way, you don't have to make 2 API calls per user-picker and you can easily control whether or not to display "me" option.

packages/ng/libraries/user/src/lib/select/input/user-select-input.component.html
```diff
@@ -16,6 +16,12 @@
 		<lu-user-paged-searcher></lu-user-paged-searcher>
 	</header>
 	<lu-user-homonyms></lu-user-homonyms>
+	<lu-option *luUserMeOption="let user" [value]="user">
```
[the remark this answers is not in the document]
lucienbertin: thats a design choice that @sanlucca and me took. we'd rather have user-select act the same everywhere even if its not super pertinent, rather than half the user-select who display me as first option, and half who dont

packages/ng/libraries/user/src/lib/select/searcher/user-searcher.model.ts
lucienbertin: i wanted the component to be self-sufficient and i'm never too shy about 2 http calls.

these calls are fired when the select opens, not when it instantiates, so thes 2 calls wont be added to the myriad that fire when a screen bootstraps, so i dont think it's _that_ impactfull

i still need to check if `me` is a possible option in the user-select with the filters that are asked so 1 of the api calls is still necessary (for example if the user-select is on the population of user you can approve expenses of)

> you can easily control

as i have stated a the previous comment i'd rather not people _control_ when it appears. either its a possible choice and it appears first, or its not and it doesn't. that way the UX for the end user is very predictible

i might add a `LU_USER_ME` injection token in the future if more component seem to need it to avoid - when provided - the first call to `/api/v3/users/me`

GuillaumeNury: 👌

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1014

PR #1016, opened 2020-05-27
[INT] remove pointer-events for disabled lu-select

RipripKipkip:


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1016

packages/ng/libraries/root/src/style/definitions/select/_select-input.scss
```diff
@@ -21,7 +21,6 @@
 
 		&[disabled] {
 			cursor: not-allowed;
-			pointer-events: none;
```
sanlucca, line 24: We just realised with Lucien that it should be the other way around. We should block pointer-events also for "is-disabled" state. Otherwise we can interact with the disabled select/textfield.

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1016

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1016

PR #1063, opened 2020-07-06
[ng] api-select - rework to force refresh on open

### LINKED ISSUE #1040: [ng] api-select - force refresh options on open
when an api-select opens, it triggers a `onOpen` that reset the page and clue. as a result if one of those two was different to the initials value (clue = `''`, page = `0`) then it retriggers an api call, which is good

_but_ if the clue and page havent been changed, the api call isn't triggerd.

this makes for some strange comportement especially when the filter has been changed while the api-seelct was closed, on open it sometimes refreshes the option to respect the new filters, sometimes it keeps the old list opf options.

to prevent that, onOpen of an api-select should force the api call to make sure the list of options is the right one

@jbosi-lucca

lucienbertin:
fixes #1040 

@vvalentin-lucca i need some stylin done because i had to change the way the loading placeholder is handled in option-picker advanced
_before_
```html
<ng-content *ngIf="!(loading$ | async); else loading"></ng-content>
<ng-template #loading>	
	<div class="lu-picker-loading loading mod-block"></div>	
</ng-template>
```

_after_
```html
<ng-content [class.is-loading]="loading$ | async"></ng-content>
```


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1063

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1063

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1063

lucienbertin:
breaks something with the user-select, dont merge yet

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1063

PR #1072, opened 2020-07-08
Fix hyphen

vvalentin-lucca:
The variable was not restored to its original state and that was causing problems.

Before:
```
.u-marginLeftSmall {
	marginleft: 1rem !important;
}
```

After:
```
.u-marginLeftSmall {
	margin-left: 1rem !important;
}
```

All the code generated is here (and is passed to the CSS validator): https://codepen.io/vincent-valentin/pen/BajxNzx?editors=0100


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1072

PR #1074, opened 2020-07-09
[FIX] Hover State for chip kill

sanlucca:
- [x] adds theming variable `components.chip`
- [x] chips doesn't have hover by default. You need to use `.mod-clickable` if you want it to be... clickable.


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1074

PR #1086, opened 2020-07-21
[SCSS][fix] white and black colors needs to be a map with color

sanlucca:
White and black "colors" wasn't used so we couldn't see it crashing.


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1086

PR #1155, opened 2020-10-07
fix(ng-formly) checkboxes and radios in mod-framed

lucienbertin:
should fix checkboxes and radios in a formly-form in a `.form.mod-framed`


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1155

PR #1157, opened 2020-10-07
fix(ng-formly) cb layout

lucienbertin:


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1157

PR #1167, opened 2020-11-10
fix(ng) ts transpil error with typescript 4

lucienbertin:
due to an override of a prop from an abstract class via a setter


GuillaumeNury (approved):

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1167

sanlucca (approved):

PR #1208, opened 2020-12-24
feat(ng-date) date-inputs - better disabling - startOn

### LINKED ISSUE #1027: [ng] date-select - open on a certain date
when the date-select opens its picker, the calendar is centered around today, it would be a plus to add an input if we want to specify a specific date around which initializing the calendar

use-case figgo
we are arond may 2020, its a restricted period until june if you want to declare a leave of absence. as a result we would like the calendar to open on july, cuz its the first month after the current restricted period

### LINKED ISSUE #1184: Disabling the calendar and its options
It should be :

- add a real disabled attribute to the calendar options outside min and max (and not just do it with a class);
- disable the navigation buttons when the months are out of scope;
- prevent the calendar from being opened from the keyboard when it is disabled.

lucienbertin:
partially fixes #1184 
fixes #1027

- use `disabled` instead of `class.is-disabled` to disable days in the calendar input when they are outside the min-max period
- add `startOn` input on calndar zand date-select comps to open the calendar on a specific date

### DIFF
--- a/packages/ng/applications/sandbox/src/app/issues/date-select-enh/date-select-enh.component.html
+++ b/packages/ng/applications/sandbox/src/app/issues/date-select-enh/date-select-enh.component.html
@@ -12,4 +12,11 @@
 	<lu-date-select class="textfield-input" [(ngModel)]="date" disabled [min]="min"></lu-date-select>
 	<span class="textfield-label">attr disabled</span>
 </label>
-<code>{{date | luDate }}</code>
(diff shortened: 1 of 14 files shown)


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1208

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1208

PR #1210, opened 2020-12-30
fix(ng-popup) max-size responsiveness

### LINKED ISSUE #1047: [ng][modal] size modificators may have bad rendering in responsive
The size modificators (.size-small, .size-medium, .size-large) that could be added to modals to alter their width are not responsive.

If viewport width < modal width, the modal lateral edges will stick to the screen edge, wich render badly.

sanlucca:
fixes #1047 
Modals are now fullscreen on mobile.

### DIFF
--- a/packages/ng/libraries/root/src/style/components/_popup.scss
+++ b/packages/ng/libraries/root/src/style/components/_popup.scss
@@ -107,7 +107,7 @@
 		border-radius: 0;
 		max-width: none;
 		max-height: 100vh;
-		width: 100vw;
+		width: 100%;
 		width: fill-available;
 		height: 100vh;
 		height: fill-available;
--- a/packages/ng/libraries/root/src/style/components/_popup.scss
+++ b/packages/ng/libraries/root/src/style/components/_popup.scss
@@ -11,7 +11,6 @@
 
 	@each $panel-size, $size-obj in _getMap("components.popup.sizes") {
 		&.mod-#{$panel-size}, &.size-#{$panel-size} {
-			max-width: none;
 			width: _component("popup.sizes."+$panel-size);
 		}
 	}
--- a/packages/ng/libraries/root/src/style/components/_popup.scss
+++ b/packages/ng/libraries/root/src/style/components/_popup.scss
@@ -105,7 +105,7 @@
 @include media_smaller_than("sm") {
 	.lu-popup-panel {
 		border-radius: 0;
-		max-width: 100vw;
+		max-width: none;
 		max-height: 100vh;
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1210

lucienbertin (approved):

packages/ng/libraries/root/src/style/components/_popup.scss
```diff
@@ -102,3 +101,22 @@
 		opacity: 0;
 	}
 }
+
+@include media_smaller_than("sm") {
+	.lu-popup-panel {
+		border-radius: 0;
+		max-width: 100vw;
```
vvalentin-lucca, line 108:
```suggestion
		max-width: none;
```

packages/ng/libraries/root/src/style/components/_popup.scss
```diff
@@ -102,3 +101,22 @@
...
+		max-height: 100vh;
+		width: 100vw;
+		width: fill-available;
```
vvalentin-lucca, line 111: Why not a 100% width and height?

sanlucca: Rather than 100vw and 100vh ? It should allow us to go fullscreen whatever the parents' style is. But might be overkill

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1210

packages/ng/libraries/root/src/style/components/_popup.scss
sanlucca: Height 100% is dangerous if the parent doesn't have a height set. I'd rather keep it to 100vh and therefore, to have a logic between the two sizes, keep width to 100vw. What do you think ?

vvalentin-lucca: I'm afraid this will cause conflicts with browsers that don't always include scrollbar calculations.

vvalentin-lucca: It all depends if we can make sure that the popins are always before the </body> taf in the DOM. Normally yes.

sanlucca: I will switch width to 100%, but might keep height to 100vh and fill-available (for chrome mobile...)

vvalentin-lucca (approved):

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1210

PR #1211, opened 2020-12-30
[fix] default palette of unselected radioButton is grey

sanlucca:


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1211

lucienbertin (approved):
good enough

radio buttons needs a good refacto imo

PR #1223, opened 2021-01-07
feat(icons) clean up how icons package works

sanlucca:
- Remove duplication of style
- Allow renaming the font
- gives access to classes generation as a mixin

### DIFF
--- a/packages/icons/src/_icons.scss
+++ b/packages/icons/src/_icons.scss
@@ -1,3 +1,4 @@
 .lucca-icon {
 	@extend %lucca-icon;
+	@include buildIconClasses;
 }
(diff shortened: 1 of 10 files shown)


packages/icons/src/_icons.scss
```diff
@@ -1,3 +1,4 @@
 .lucca-icon {
 	@extend %lucca-icon;
+	@include buildIconClasses;
```
BertrandPodevin, line 3: Here the selector `.lucca-icon` is repeated for each icon in the final dist, and we have ~200 icons, wich leads to somewhat wasted dist size. Could be optimized, but it's a rather marginal gain

BertrandPodevin (approved):

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1223

PR #1244, opened 2021-02-22
feat(scss) u-listReset for dl

vvalentin-lucca:
https://codepen.io/vincent-valentin/pen/KKNXojm

### DIFF
--- a/packages/scss/src/mixins/_utilities.scss
+++ b/packages/scss/src/mixins/_utilities.scss
@@ -20,8 +20,12 @@
 
 @mixin listReset($suffix: '') {
 	list-style-type: none #{$suffix};
-	margin: 0 #{$suffix};
 	padding: 0 #{$suffix};
+	
+	&,
+	> * {
+		margin: 0 #{$suffix};
+	}
 }
 
 @mixin buttonReset($suffix: '') {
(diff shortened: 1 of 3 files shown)


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1244

packages/scss/src/mixins/_utilities.scss
```diff
@@ -20,8 +20,12 @@
 
 @mixin listReset($suffix: '') {
 	list-style-type: none #{$suffix};
-	margin: 0 #{$suffix};
 	padding: 0 #{$suffix};
+	
+	&,
+	> * {
```
BertrandPodevin, line 26: I can seesome potentials side effects to add `> *` in this reset. I think it would be safer to juste create a specific reset for dl dt dd lists

vvalentin-lucca: You're right! At least for the `div` nested in the `dl`.

How about a `&, dd { margin: 0; }` ?

BertrandPodevin: There are 2 way to do it :
- As you suggest, just add `&, dt, dd { margin: 0; }` in the reset
- Or create a new `u-definitionListReset` with
```
@include listReset;
&, > * { margin: 0; }
```
i lean toward the second option, since list and definition list are slightly different DOM wise.

LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1244

sanlucca (approved):

PR #1263, opened 2021-03-25
fix(ng) establishment-picker selects the same value more than once

bastienguillon:
Fixes a bug where the same value was selected more than once by the establishment picker

[HUB-3287]

[HUB-3287]: https://luccasoftware.atlassian.net/browse/HUB-3287


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1263

lucienbertin (approved):

PR #1284, opened 2021-04-16
[PoC] storybook for package ng

lucienbertin:
adds a storybook portal to pacakge ng

still a bit buggy but good enough for now


lucienbertin:
not the implementation chosen, see #1286

PR #1304, opened 2021-05-04
tech(all) Volta and clean up peer deps

lucienbertin:
use volta to pilot which node/npm version to use
clean up peer deps, the rule now is all lf/xxx should be on the same major


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1304

PR #1327, opened 2021-05-25
WIP add container for sticky column drop shadow on scroll

BertrandPodevin:
Add pure css column sticky shadow on scroll for table.mod-columnSticky

image

### DIFF
--- a/packages/scss/src/components/_table.scss
+++ b/packages/scss/src/components/_table.scss
@@ -538,8 +538,10 @@
 }
 
 
-// Fixed colmuns width
-$nb-max-columns: 12;
+// Fixed columns width
+$min-col-width: 1; //value in rem
+$max-col-width: 30; //value in rem
+
 .table {
 	&.mod-layoutFixed {
 		table-layout: fixed;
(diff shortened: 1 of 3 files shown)


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1327

packages/scss/src/components/_table.scss
```diff
@@ -621,4 +621,47 @@ $max-col-width: 20; //value in rem
 			right: $i * 1rem;
 		}
 	}
+}
+
+//columns sticky css drop shadow on scroll
+.tableColumnStickyWrapper {
+	$columnStickyShadow-backgroundSize-width: 10px;
```
sanlucca, line 628: Use our var system rather than custom vars

packages/scss/src/components/_table.scss
```diff
@@ -621,4 +621,47 @@ $max-col-width: 20; //value in rem
...
+	background-image: 
+		/* Shadows */ 
+		linear-gradient(to right, _color(white), _color(white)),
+		linear-gradient(to right, _color(white), _color(white)),
+		/* Shadow covers */ 
+		linear-gradient(to right, rgba(0,0,0,.25), rgba(255,255,255,0)),
+		linear-gradient(to left, rgba(0,0,0,.25), rgba(255,255,255,0));   
+
+	background-position:
+		var(--columnStickyShadow-leftOffset) center,
```
sanlucca, line 640: Don't use css-vars directly. prefer using our theming system. It will be able to switch from hard values to css-vars easily with a single bool.

sanlucca (changes requested):

Issue #1330, opened 2021-05-26
[DS to LF] Palette primary, secondary


AlexiaAlarza:
Modify the primary and secondary colors to identify them correctly (hover, active...). 
Choose the main colors : https://www.figma.com/file/sO3msermeMuyFvUAcXrbn5/Main-Components-2.0?node-id=3%3A27

![Capture d’écran 2021-05-26 à 15 41 29](https://user-images.githubusercontent.com/77682210/119670050-dbcca500-be38-11eb-817f-c5a1ccac32bc.png)

PR #1375, opened 2021-05-28
[DS to LF] Palette primary, secondary

BertrandPodevin:
fix for https://github.com/LuccaSA/lucca-front/issues/1330


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1375

sanlucca:
We can't do it right now for a simple reason : our ng components can't be themed on IE11 (cssvars are mandatory to theme ng components). 
That's why LF palettes are grey for now, as it makes it not painful to watch in any software on IE11. It can be done after August, but not before. 
Maybe we can find a way to have this palette in the SCSS package and an override in the default ng package, but we need to be careful.

PR #1394, opened 2021-06-24
feat(ng-a11y) popover triggers

### LINKED ISSUE #1393: Lupopover - focus and hover
Allow focus + hover on LuPopover component

lucienbertin:
fixes #1393 

popover triggers `hover` and `focus` now act the same because js events focus and hover should be understood as 2 ways to represent the same end user interaction with the element

so a popover with trigger `hover` will open when the trigger is focused, and a popover with the `focus` trigger will appear when hovered


LuccaIntegration:
jenkins auto deploy http://lucca-front.lucca.local/PR-1394

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1394

PR #1399, opened 2021-07-05
fix(ng-user) me-first directive wasnt applying the filters of the user-select

lucienbertin:
it wont work when one of the filter applied is `id=notequal,1,2,3`, that just breaks the user-select


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1399

shajz (approved):

PR #1428, opened 2021-08-04
fix(ng/material) use v12 syntax for @angular/material theming

GuillaumeNury:
LF overrides @angular/material typography using [v11 syntax](https://v11.material.angular.io/guide/typography).

This PR uses [v12+ syntax](https://material.angular.io/guide/typography).

As LF has to be compatible with Angular v11 and v12, two new SCSS entry point have been added. When using Angular 11 with LF 7 : 
```scss
/* Before */
@import "~@lucca-front/ng/material/style/main.overridable.scss";

/* After*/
@import "~@lucca-front/ng/material/style/main-v11.overridable.scss";
```

PR #1438, opened 2021-08-25
[NG][FIX] Stop importing all of material CSS

sanlucca:
PR #1428 used `mat.core` mixin, importing all the style from material. We replace it with the mixin `mat.all-component-typographies()` which is the correct update of the deprecated code.


BertrandPodevin (approved):

vvalentin-lucca (approved):

shajz (approved):
👏

PR #1453, opened 2021-09-29
feat(ng/user) MeOption improvements

### LINKED ISSUE #1447: user-select : Hide myself when search
In the user-select picker, "me" is always displayed as the top value. I understand it is useful and consistent to keep it by default, but when you start to search for someone, and your name does not match the search clue, it does not make sense to keep it in the options.

GuillaumeNury:
## Description

* Stop displaying "me" option when clue is not empty (fix #1447)
* Do not display user twice (when current user is in options)

Before :

image

After :

image (6)

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1453

RipripKipkip (approved):

PR #1466, opened 2021-10-20
Fix appearance for input with iOS

vvalentin-lucca:
Without that, their borders don't appear.

Capture d’écran 2021-10-20 à 09 57 48

Capture d’écran 2021-10-20 à 09 58 37


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1466

PR #1468, opened 2021-10-21
Fix menu legacy regression

vvalentin-lucca:


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1468

PR #1490, opened 2021-12-03
feat(scss) use grey palette for disabled state on framed form components

MarcFairbrother:
## Description

- uses grey palette for disabled state on framed form components
- adds story for framed forms

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1490

packages/scss/src/mixins/_forms.scss
```diff
@@ -1,33 +1,35 @@
 @import "../mixins/utilities";
 
 @mixin fieldState($fieldname, $state) {
-	background-color: _color($state, "50");
+	&:not(:disabled) {
```
AlexiaAlarza, line 4: &:not(:disabled):not(.is-disabled) à rajouter ?

packages/scss/src/components/inputs/textfield/_textfield.framed.scss
```diff
@@ -60,74 +60,23 @@
...
-							}
-						}
-					}
-				}
-			}
-
-			&.is-error {
-				box-shadow: fakeborderoverlay(_color("error"));
-				z-index: 3;
+		&:not(:disabled, .is-error, .is-success, .is-valid, .is-warning, .is-invalid, .ng-invalid) {
```
AlexiaAlarza, line 67: &:not(:disabled):not(.is-disabled) à rajouter ?

packages/scss/src/components/_form.scss
```diff
@@ -139,7 +140,7 @@
 		clear: both;
 	}
 
-	&:hover {
+	&:not(.is-disabled):hover {
```
AlexiaAlarza, line 143: &:not(:disabled):not(.is-disabled) à rajouter ?

MarcFairbrother: This `.is-disabled` class is used on a parent of the `<input>`, not on the input itself so we cannot target the `disabled` attribute.

packages/scss/src/components/inputs/textfield/_textfield.framed.scss
MarcFairbrother: The `.is-disabled` class doesn't work on the framed form elements (styles are overridden by the more specific framed style declarations). I feel like this is a bad practice in any case and the `.is-disabled` class should be deprecated in favour of the `disabled` attribute.

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1490

AlexiaAlarza (approved):

PR #1495, opened 2021-12-20
FIX(scss) Hover states for fields in mod-framed

vvalentin-lucca:
The grid background on the hover was not in accordance with the design system.
It is replaced by a more pronounced border color. 

A white background is also added to override the inherited background color of the fields outside the mod-framed.

-----

Capture d’écran 2021-12-20 à 17 18 21

Capture d’écran 2021-12-20 à 17 18 42

### DIFF
--- a/packages/scss/src/components/_form.scss
+++ b/packages/scss/src/components/_form.scss
@@ -140,14 +140,19 @@
 		clear: both;
 	}
 
-	&:not(.is-disabled):hover {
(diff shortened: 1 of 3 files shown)


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1495

MarcFairbrother (approved):

PR #1461, opened 2021-10-06
Responsive for boutons grouped

vvalentin-lucca:
The grouped buttons were not responsive.

- They can now adapt to the width of the viewport, but also be easily overloaded with media-queries or classes (`.u-flexDirectionColumn` for example).
- The dark border that used to separate the buttons has been replaced by a light border.
- The code also provides for the presence of a unordered list but this remains optional so as not to bring regression.
- The story has also been updated to showcase this mod.

-----

Capture d’écran 2022-02-21 à 11 10 40

Capture d’écran 2022-02-22 à 18 22 10

PR #1544, opened 2022-02-25
Responsive for boutons grouped

vvalentin-lucca:
The grouped buttons were not responsive.

- They can now adapt to the width of the viewport, but also be easily overloaded with media-queries or classes (`.u-flexDirectionColumn` for example).
- The dark border that used to separate the buttons has been replaced by a light border.
- The code also provides for the presence of a unordered list but this remains optional so as not to bring regression.
- The story has also been updated to showcase this mod.

-----

Capture d’écran 2022-02-21 à 11 10 40

Capture d’écran 2022-02-22 à 18 22 10


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1544

vvalentin-lucca:
New version of this PR (#1461) (but with RC and not master).

BertrandPodevin (approved):

PR #1585, opened 2022-04-05
[Font size] fix standard

jeremie-lucca:
Standard size is used as H4 and Body 1 (which lately diverged in Figma). 
Reducing the standard line height on LF impacted the height of multiple components (buttons, selects...)

For now we are reverting the standard line height but we should have 2 specifics values in the future.


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1585

vvalentin-lucca (approved):

PR #1627, opened 2022-04-22
[Button groups] Separator color

jeremie-lucca:
image
image
image
image

### DIFF
--- a/packages/scss/src/components/_button-group.scss
+++ b/packages/scss/src/components/_button-group.scss
@@ -3,7 +3,6 @@
 	list-style-type: none;
 	padding: 0;
 	margin: 0;
-	gap: _theme('commons.divider.width');
 	border-radius: _component('button.border.radius');
 	overflow: hidden;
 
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1627

packages/scss/src/components/_button-group.scss
```diff
@@ -3,7 +3,6 @@
 	list-style-type: none;
 	padding: 0;
 	margin: 0;
-	gap: _theme('commons.divider.width');
```
BertrandPodevin, line 6: i think the use of gap was here was added with responsive consideration in mind. Maybe we could just keep it and change the background color of `button-group` ? That should do the trick

vvalentin-lucca: Exactly!

And from memory we had been asked to the design if we could switch to white background.
It's more like LF2DS here. 😉

PR #1642, opened 2022-05-02
[Button] Size small height

jeremie-lucca:
36px > 32px

### DIFF
--- a/packages/scss/src/theming/components/_button.theme.scss
+++ b/packages/scss/src/theming/components/_button.theme.scss
@@ -7,7 +7,7 @@ $button: (
 	padding: _theme('spacings.smaller') _theme('spacings.standard'),
 	small: (
 		font-size: _theme('sizes.small.font-size'),
-		line-height: _theme('sizes.small.line-height'),
+		line-height: _theme('sizes.smaller.line-height'),
 		padding: _theme('spacings.smaller') _theme('spacings.small'),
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1642

packages/scss/src/theming/components/_button.theme.scss
```diff
@@ -7,7 +7,7 @@ $button: (
 	padding: _theme('spacings.smaller') _theme('spacings.standard'),
 	small: (
 		font-size: _theme('sizes.small.font-size'),
-		line-height: _theme('sizes.small.line-height'),
+		line-height: _theme('sizes.smaller.line-height'),
```
MarcFairbrother, line 10: The small and smaller line-heights used to have the same value. Are we sure that we want them to have different values now?

jeremie-lucca: That's what we already do on .smaller : 
image
But to be honest I'm not even sure the theming should be linked to size variables!

MarcFairbrother: Ok, let's use this fix for now. It's a bit dodgy to use different size names so we should think about a more consistent solution for the future.

MarcFairbrother (approved):

PR #1665, opened 2022-05-13
ref: no more mod-zeroHeight

GuillaumeNury:
## Description
Adding `mod-zeroHeight` in storybook urls in no longer needed as the new system detect "parent" iframe to decide whether or not custom style is necessary. This custom style is added inside head element so it is loaded sooner thant before (to avoid UI flashing).

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1665

shajz (approved):

PR #1758, opened 2022-07-11
Fix order declaration

vvalentin-lucca:
The order of the utilities is important.
Here I quickly try to reverse them, but perhaps it will be necessary to copy the order of LF 9 precisely.


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1758/storybook

PR #1856, opened 2022-09-14
overlayRef.attached is not a condition to prevent closePopover

guillerot:


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1856/storybook

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1856/storybook

shajz (dismissed):

packages/ng/popover/src/lib/trigger/popover-trigger.model.ts
```diff
@@ -199,8 +199,10 @@ export abstract class ALuPopoverTrigger<TPanel extends ILuPopoverPanel = ILuPopo
 
 	/** Closes the popover. */
 	closePopover(): void {
-		if (this._overlayRef && this._overlayRef.hasAttached()) {
-			this._overlayRef.detach();
+		if (this._overlayRef) {
```
GuillaumeNury, line 202: should we call `emitClose` when we have an unattached overlayRef ?

GuillaumeNury: > Maybe there is a way to call emitClose sooner (when the "real" close occurs with the scrolling)

I would prefer that.

In the `_createOverlay` method, we could use something like `this._overlayRef.detachments().subscribe(...)` to detect popover closing.
But with this, we have to make sure that close event is not emitted twice in the "normal" use-case.

guillerot: @GuillaumeNury i've tried to take advantage of `_overlayRef.detachments`
See my last proposal

GuillaumeNury (approved):
Thx @guillerot

shajz (approved):

PR #1860, opened 2022-09-15
fontWeight

### LINKED ISSUE #1813: Utility .u-fontWeightRegular is missing
Also, to discuss :
Should we remove `u-fontWeight400/600` and only keep `u-fontWeightRegular/Bold`?

vvalentin-lucca:
Fix #1813 

Add selectors font font weights.

(And remove duplicate seletor.)

### DIFF
--- a/packages/scss/src/components/util/index.scss
+++ b/packages/scss/src/components/util/index.scss
@@ -127,10 +127,6 @@
 	color: var(--palettes-grey-700) !important;
 }
 
-.u-ellipsis {
-	@include text.ellipsis;
-}
-
 .u-unit {
 	font-size: 0.66em !important;
 }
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1860/storybook

packages/scss/src/components/util/index.scss
```diff
@@ -147,6 +143,14 @@
 	@include text.ellipsis('!important');
 }
 
+.u-fontWeightRegular {
```
BertrandPodevin, line 146: Since `font-weight` properties are already listed in core, it seem unconsistant to find those new here, even if their syntax is slightly different.

Maybe we could do something like  
`$fontWeight: '200', '300', '400', '500', '600', '700', '800', '900', 'normal', 'bold', 'regular:400', 'semiBold:600';` in core and add some logic in `core.classes` ? That doesn't sound too complex

BertrandPodevin (approved):

packages/scss/src/components/util/index.scss
vvalentin-lucca: My opinion here: https://github.com/LuccaSA/lucca-front/issues/1813#issuecomment-1237944561

PR #1922, opened 2022-11-29
Update config with default parameters

vvalentin-lucca:
## Description

By adding the keyword "default", we allow overloads on the colours in the products.

It will be useful to apply the new colour charts progressively in the products before doing it again here in a global way.

-----

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1922/storybook

jeremie-lucca (approved):

PR #1923, opened 2022-12-02
[Pagination] Add compact story

jeremie-lucca:
## Description

Add compact story

-----

image

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1923/storybook

vvalentin-lucca:
Wouldn't that be better served as a parameter to the original story rather than a new story?

PR #1939, opened 2022-12-21
[DS to LF] Messages : Toast 

RipripKipkip:
## Description

- [x] Add palettes
- [x]  Put the title in block to make the content go underneath
- [x]  Plan with an icon
- [x]  Change default background + palette to 800
- [ ]  Fix toasts-item-kill position. Should be right - top, not right - center

### Dev
- [ ] Add mod-icon only when there is an icon

-----

Before
image

After
image
image

-----

### DIFF
--- a/packages/ng/toast/src/toasts.component.html
+++ b/packages/ng/toast/src/toasts.component.html
@@ -7,7 +7,8 @@
 			[ngClass]="iconClassByToastType[toast.type]"
 		>
 		</span>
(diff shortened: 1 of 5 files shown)


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1939/storybook

jeremie-lucca:
Already fixed

PR #1953, opened 2023-01-05
fix(sidepanel): Sidepanel where broken because of the new inject syntax

GuillaumeNury:
## Description

🐛 Sidepanel where broken because of the new inject syntax: we need to provide all inherited services (even if not necessary...)

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-1953/storybook

F2X (approved):

PR #2007, opened 2023-02-23
fix(schematic): CRLF/LF issue on some projects

GuillaumeNury:
## Description

When a file use CRLF line endings (hey Windows !), the replaced content is broken because typescript normalize line endings with only LF.

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2007/storybook

shajz (approved):

PR #2018, opened 2023-02-27
[Forms] Fix variables

jeremie-lucca:
## Description

Fix variables on `.form` scope

-----

Not a perfect solution as we duplicated variables, but it avoids breakings before the form/field refactor. 

-----

### DIFF
--- a/packages/scss/src/components/form/index.scss
+++ b/packages/scss/src/components/form/index.scss
@@ -3,6 +3,7 @@
 @use 'exports' as *;
 
 .form {
+	@include vars;
 }
 
 .form-group {


vvalentin-lucca (approved):

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2018/storybook

packages/scss/src/components/form/index.scss
```diff
@@ -3,6 +3,7 @@
 @use 'exports' as *;
 
 .form {
```
BertrandPodevin, line 5: You can slightly optimize the whole thing with :
```
.form,
.form-group {
   @include vars;
}
```

jeremie-lucca: True, but it would lead to have multiple form-group declarations.
Let's improve the whole thing during the next refactor :)

PR #2119, opened 2023-05-12
Font family inherit for button and forms

vvalentin-lucca:
## Description

Rather than declare the used family font several times, we take advantage of the inheritance when possible.

-----

This will make it easier to override the font when necessary, with one call instead of several.

-----


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2119/storybook

jeremie-lucca (approved):

PR #2146, opened 2023-06-19
[Components v3] Card 

jeremie-lucca:
## Description

Update card UI

-----

image

-----

### DIFF
--- a/packages/scss/src/components/card/component.scss
+++ b/packages/scss/src/components/card/component.scss
@@ -3,9 +3,11 @@
 	border-radius: var(--commons-borderRadius-L);
 	border-color: var(--commons-divider-color);
 	border-width: 1px;
+	display: block;
 	margin-bottom: var(--spacings-S);
 	border-style: solid;
-	display: block;
+	flex-direction: row;
+	row-gap: var(--spacings-S);
 	position: relative;
 	transition-duration: var(--commons-animations-durations-fast);
 	transition-property: background-color, box-shadow, color;
@@ -13,6 +15,20 @@
 	@at-root ($atRoot) {
 		.card-content {
 			padding: var(--components-card-content-padding);
+		}
+
+		.card-title {
+			padding: 0;
+			margin-bottom: var(--spacings-XS);
+		}
+
+		.card-action {
+			display: flex;
+			align-items: center;
+			justify-content: center;
+			border-left: 1px solid var(--commons-border-200);
+			padding-left: var(--spacings-S);
+			margin-left: var(--spacings-S);
 		}
 
 		.card-footer {
--- a/packages/scss/src/components/card/mods.scss
+++ b/packages/scss/src/components/card/mods.scss
@@ -7,7 +7,7 @@
 	text-decoration: none;
 	cursor: pointer;
 
-	&:focus  {
+	&:focus {
 		outline: none;
 	}
 
@@ -20,13 +20,17 @@
 }
 
 @mixin elevated {
-	box-shadow: var(--commons-elevations-elevation-1);
+	box-shadow: var(--commons-boxShadow-S);
 	border: 0;
 }
 
-@mixin clickableElevated {
-	&:focus-visible,
-	&:hover {
-		box-shadow: var(--commons-elevations-elevation-3);
+@mixin action {
+	display: flex;
+	flex-direction: row;
+	padding: var(--components-card-content-padding);
+
+	.card-content {
+		flex: 1;
+		padding: 0;
 	}
 }
--- a/packages/scss/src/components/card/mods.scss
+++ b/packages/scss/src/components/card/mods.scss
@@ -2,10 +2,6 @@
 	--components-card-background: var(--palettes-grey-50);
 }
 
-@mixin action {
-	opacity: .5;
-}
-
 @mixin clickable {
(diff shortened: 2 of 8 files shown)


LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2146/storybook

stories/documentation/structure/cards/cards-action.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+			<p>Contenu de la carte</p>
+		</div>
+		<div class="card-action">
+			<button type="button" class="actionIcon" luTooltip="Modifier">
+			  <span aria-hidden="true" class="lucca-icon icon-arrowFullEast"></span>
+			  <span class="u-mask">Accéder</span>
+			</button>
+		</div>
+	</div>
+	`
```
MarcFairbrother, line 31: Design will probably ask for the entire card to be clickable. A frquent use case we have is that these cards act as links to content. I think we will need this component to handle the `<a class="card mod-action">` use case.

jeremie-lucca: After a chat with @vvalentin-lucca it seems that having a whole card as link element can lead to accessibility issue. 
It seems that the best way is to let Angular handle the click : https://codepen.io/matuzo/full/ZEqbXOM
I'll add a `mod-action `anyway to cover hover effect.

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2146/storybook

packages/scss/src/components/card/component.scss
```diff
@@ -3,9 +3,11 @@
 	border-radius: var(--commons-borderRadius-L);
 	border-color: var(--commons-divider-color);
 	border-width: 1px;
+	display: block;
```
vvalentin-lucca, line 6: `display: flex` ?

packages/scss/src/components/card/component.scss
```diff
@@ -3,9 +3,11 @@
 	border-radius: var(--commons-borderRadius-L);
 	border-color: var(--commons-divider-color);
 	border-width: 1px;
```
MarcFairbrother, line 5: Why not use a box-shadow rather than a border?

packages/scss/src/components/card/mods.scss
```diff
@@ -2,12 +2,16 @@
 	--components-card-background: var(--palettes-grey-50);
 }
 
+@mixin action {
+	opacity: .5;
```
MarcFairbrother, line 6: Shouldn't we use a light version of the palette color instead of opacity?

packages/scss/src/components/card/mods.scss
```diff
@@ -20,13 +24,24 @@
...
 }
 
-@mixin clickableElevated {
-	&:focus-visible,
+@mixin action {
+	box-shadow: var(--commons-boxShadow-S);
+	border: 0;
+	cursor: pointer;
+	display: flex;
+	flex-direction: row;
```
MarcFairbrother, line 36: Row is the default value, do we need to define the property?

packages/scss/src/components/card/vars.scss
```diff
@@ -1,5 +1,5 @@
 @mixin vars {
-	--components-card-content-padding: var(--spacings-M);
+	--components-card-content-padding: .75rem var(--spacings-S);
 	--components-card-footer-padding: 0.33rem var(--spacings-M);
```
MarcFairbrother, line 3: 0.33 rem?

packages/scss/src/components/card/component.scss
jeremie-lucca: It would have impact on cards with `card-content` and `card-footer`.  It would be doable by adding a `flex-direction`, but I a bit afraid of side effects.

packages/scss/src/components/card/component.scss
jeremie-lucca: On this project, a large amount of components needed to be updated, so I edited only the essential to limit the risks. As border property is working fine and might be overrided on some cases, it seems safer to keep it as it is for now.

packages/scss/src/components/card/mods.scss
jeremie-lucca: It was only a test that I forget to remove. 🤦

packages/scss/src/components/card/mods.scss
jeremie-lucca: Nop, it's related to my first approach where the default card has a `flex-direction: column`.
Fixed.

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2146/storybook

packages/scss/src/components/card/vars.scss
jeremie-lucca: Old code for a mod not mentioned on Figma.
Fixed anyway 🙂

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2146/storybook

MarcFairbrother (approved):

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2146/storybook

PR #2148, opened 2023-06-21
[Tooltip] Add luTooltipAsHtml input

belabb-z:
## Description

Add a `luTooltipAsHtml` input for the tooltip component to allow the content to be interpreted as HTML.

-----

The content of the tooltip is now passed in a mustache by default, except when the `luTooltipAsHtml` input is set to `true`.

-----

### DIFF
--- a/packages/ng/tooltip/src/lib/panel/tooltip-panel.component.html
+++ b/packages/ng/tooltip/src/lib/panel/tooltip-panel.component.html
@@ -7,5 +7,8 @@
 	[attr.id]="panelId"
 	[@transformTooltip]="'enter'"
 >
-	<div class="lu-tooltip-content" [ngClass]="contentClassesMap" [innerHtml]="content"></div>
+	<div *ngIf="contentAsHtml; else contentAsString" class="lu-tooltip-content" [ngClass]="contentClassesMap" [innerHtml]="content"></div>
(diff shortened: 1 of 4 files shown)


g-drouard:
Ça fait un sacré breaking changes si on met à `false` la valeur par défaut

belabb-z:
> Ça fait un sacré breaking change si on met à `false` la valeur par défaut du `luTooltipAsHtml` mais c'est peut-être ce qu'on veut.

en effet, c'est breaking mais ça permet aussi de faire le tour des usages

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2148/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/4262/ui-diff)

MarcFairbrother:
The guideline states that tooltips shouldn't contain HTML. What's the use case we're trying to address?

https://prisme.lucca.io/94310e217/p/9285d2-tooltip

MarcFairbrother:
Maybe the guideline needs to be more precise. @jeremie-lucca was the intention not to be able to include links inside a tooltip? It could be useful to allow HTML to handle add styles in some cases.

image

vvalentin-lucca:
But isn't that already addressed by lu-popover?

PR #2166, opened 2023-07-13
feat(callout): new callout wrapper

Supamiu:
## Description

First implementation of the callout wrapper, aiming at wrapping the callout structure in an Angular component for better maintainability.

This PR also includes new stories using the new wrapper, as well as using the new CSF3 Storybook system to make it easier to share code between various stories.

-----

Aside from the implementation itself, something that needs to be reviewed is if we want the `hidden` event to hide the callout too (so have the callout `*ngIf` its own template) or not.

One of the main pros is how easy it'd be to use, one of the main cons is that you'll probably want to store this "hidden" flag somewhere like in localstorage or further, and having the component handle it would make it too smart for what it is. Not to mention the fact that it also means adding a `ng-container` to wrap everything while a parent component that handles hidden would just `*ngIf` the callout.

Another approach could be to make it a two-way binding and make it hide itself upon clicking on the X icon, letting the parent component decide whether they want to store that info or not.

-----

### DIFF
--- a/packages/ng/callout/src/lib/callout/callout.component.html
+++ b/packages/ng/callout/src/lib/callout/callout.component.html
@@ -1,0 +1,12 @@
+<div class="callout palette-{{palette}}" [class.mod-S]="size === 's'" [class.mod-tiny]="tiny">
+	<div class="callout-icon" *ngIf="icon">
+		<span aria-hidden="true" class="lucca-icon icon-{{icon}}"></span>
+	</div>
+	<div class="callout-content">
+		<strong class="callout-content-title" *ngIf="title">{{title}}</strong>
+		<div class="callout-content-description">
+			<ng-content></ng-content>
+		</div>
+	</div>
+	<button *ngIf="removable" type="button" class="callout-kill" (click)="hide()"></button>
+</div>
--- a/packages/ng/callout/src/lib/callout/callout.component.scss
+++ b/packages/ng/callout/src/lib/callout/callout.component.scss
@@ -1,0 +1,1 @@
+@use '@lucca-front/scss/src/components/callout';
--- a/packages/ng/callout/src/lib/callout/callout.component.ts
+++ b/packages/ng/callout/src/lib/callout/callout.component.ts
@@ -1,0 +1,71 @@
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { Palette, SizeMS } from '@lucca-front/ng/core';
+
+@Component({
+	selector: 'lu-callout',
+	standalone: true,
+	imports: [CommonModule],
+	templateUrl: './callout.component.html',
+	styleUrls: ['./callout.component.scss'],
+})
+export class CalloutComponent {
+	@Input()
+	/**
+	 * The title of the callout
+	 */
+	title: string;
+
+	@Input()
+	/**
+	 * Which palette should be used for the entire callout.
+	 * Defaults to none (inherits parent palette)
+	 */
+	palette: Palette = 'none';
+
+	@Input()
+	/**
+	 * Which size should the callout be? Defaults to medium
+	 */
+	size: SizeMS = 'm';
+
+	@Input()
+	/**
+	 * Should we display the remove icon?
+	 *
+	 * IMPORTANT: the callout won't hide itself, you're responsible for removing
+	 * it using *ngIf as you might want to store the information that it has been
+	 * hidden. Hook on the `hidden` event emitter for that.
+	 */
+	removable = false;
+
+	@Input()
+	/**
+	 * Which icon should we display in the callout if any?
+	 * Defaults to no icon.
+	 */
+	icon: 'info' | 'success' | 'warning' | 'error' | 'help' | string;
+
+	@Input()
+	/**
+	 * Should we use tiny mode?
+	 * WARNING: tiny mode should only be used without a title, there's no runtime
+	 * check for this for performance reasons but make sure to never have both title
+	 * and tiny.
+	 */
+	tiny: boolean;
+
+	@Output()
+	/**
+	 * Emits void when the callout's close button is clicked.
+	 *
+	 */
+	hidden: EventEmitter<void> = new EventEmitter<void>();
+
+	/**
+	 * Emits the hidden event for the consumer to hide the callout.
+	 */
+	hide(): void {
+		this.hidden.emit();
+	}
+}
--- a/packages/ng/callout/src/lib/index.ts
+++ b/packages/ng/callout/src/lib/index.ts
@@ -1,0 +1,1 @@
+export * from './callout/callout.component';
--- a/packages/ng/core/src/lib/type/style.ts
+++ b/packages/ng/core/src/lib/type/style.ts
@@ -1,0 +1,9 @@
+/**
+ * Available CSS palettes
+ */
+export type Palette = 'success' | 'warning' | 'error' | 'primary' | 'grey' | 'none';
+
+/**
+ * Size that accepts only medium or small
+ */
+export type SizeMS = 'm' | 's';
--- a/stories/documentation/feedback/callout/callout-tiny.stories.ts
+++ b/stories/documentation/feedback/callout/callout-tiny.stories.ts
@@ -1,70 +1,52 @@
-import { Meta, Story } from '@storybook/angular';
-
-interface CalloutTinyStory {
-	s: boolean;
-	palette: string;
-	icon: string;
-}
+import { Meta, StoryObj } from '@storybook/angular';
+import { CalloutComponent } from '@lucca-front/ng/callout';
+import { default as BasicStory } from './callout-basic.stories';
 
 export default {
-	title: 'Documentation/Feedback/Callout/Tiny',
-	argTypes: {
-		s: {
-			control: {
-				type: 'boolean',
-			},
-			description: "Taille : Small",
-		},
-		palette: {
-			options: ['', 'palette-success', 'palette-warning', 'palette-error'],
-			control: {
-				type: 'select',
-			},
-		},
-		icon: {
-			options: ['icon-help', 'icon-success', 'icon-warning', 'icon-error'],
-			control: {
-				type: 'select',
-			},
-		},
-	},
+	...BasicStory,
+	title: 'Documentation/Feedback/Callout/Tiny'
 } as Meta;
 
-function getTemplate(args: CalloutTinyStory): string {
-	const s = args.s ? `mod-S` : '';
-	const palette = args.palette;
-	const icon = args.icon;
-	let text: {title: string, description: string};
-	switch (args.palette) {
-		case 'palette-success':
-				text = {title: 'Cool!', description: 'Je suis un callout de succès :)'};
-				break;
-		case 'palette-warning':
-				text = {title: 'Hmmm...', description: 'Je suis un callout d\'alarme :|'};
-				break;
-		case 'palette-error':
-				text = {title: 'Oops!', description: 'Je suis un callout d\'erreur :('};
-				break;
-		default:
-			text = {title: 'Besoin d\'aide ?', description: 'Je suis un callout standard'};
-			break;
-  };
-	return `
-	<div class="callout mod-tiny ${s} ${palette}">
-		<div class="callout-icon">
-			<span aria-hidden="true" class="lucca-icon ${icon}"></span>
-		</div>
-		<div class="callout-content">
-			99
-		</div>
-	</div>
-	`
-}
-
-const Template: Story<CalloutTinyStory> = (args: CalloutTinyStory) => ({
-	props: args,
-	template: getTemplate(args),
-});
-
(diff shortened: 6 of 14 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/4536/ui-diff)

packages/ng/callout/src/lib/callout/callout.component.html
```diff
@@ -0,0 +1,12 @@
...
+	<div class="callout-icon" *ngIf="icon">
+		<span aria-hidden="true" class="lucca-icon icon-{{icon}}"></span>
+	</div>
+	<div class="callout-content">
+		<strong class="callout-content-title" *ngIf="title">{{title}}</strong>
+		<div class="callout-content-description">
+			<ng-content></ng-content>
+		</div>
+	</div>
+	<button *ngIf="removable" type="button" class="callout-kill" (click)="hide()"></button>
```
MarcFairbrother, line 11: The button is missing a masked text for accessibility.

```suggestion
	<button *ngIf="removable" type="button" class="callout-kill" (click)="hide()">
	    <span class="u-mask">{{ A11Y_DISMISS_CALLOUT | translate}}</span>
	</button>
```

packages/ng/callout/src/lib/callout/callout.component.html
```diff
@@ -0,0 +1,12 @@
+<div class="callout palette-{{palette}}" [class.mod-S]="size === 's'" [class.mod-tiny]="tiny">
```
MarcFairbrother, line 1: `s` and `tiny` could both be values of a `size` input.

packages/ng/callout/src/lib/callout/callout.component.ts
```diff
@@ -0,0 +1,71 @@
...
+	 * Which palette should be used for the entire callout.
+	 * Defaults to none (inherits parent palette)
+	 */
+	palette: Palette = 'none';
+
+	@Input()
+	/**
+	 * Which size should the callout be? Defaults to medium
+	 */
+	size: SizeMS = 'm';
```
vvalentin-lucca, line 30: With the possibilities in the variable name, won't it be annoying in the future if we add more?

packages/ng/callout/src/lib/callout/callout.component.ts
```diff
@@ -0,0 +1,71 @@
...
+import { CommonModule } from '@angular/common';
+import { Palette, SizeMS } from '@lucca-front/ng/core';
+
+@Component({
+	selector: 'lu-callout',
+	standalone: true,
+	imports: [CommonModule],
+	templateUrl: './callout.component.html',
+	styleUrls: ['./callout.component.scss'],
+})
```
jbosi, line 11: `changeDetection: ChangeDetectionStrategy.OnPush` à rajouter ?

packages/ng/callout/src/lib/callout/callout.component.html
vvalentin-lucca: @MarcFairbrother 

FYI:
- `tiny` switches the component to `inline-block` and reduces the `gap`;
- `size` changes the component size otherwise.

packages/ng/callout/src/lib/callout/callout.component.ts
Supamiu: I think that ideally, we'd have to define a list of size names and use them everywhere to have more consistency tbh.
This is more of an example one, but I agree that we should just use the same names everywhere and have types to define which ones we want to allow where (for instance, have a type that's `SizeSML` that is defined by `'small' | 'medium' | 'large'`, one `SizeSM` that's the same without large, etc.

I don't know if this naming change should be done starting now or not tho.

packages/ng/callout/src/lib/callout/callout.component.html
Supamiu: Yeah I think this is more of a design-related decision (because it also depends on figma mappings)

packages/ng/callout/src/lib/callout/callout.component.html
Supamiu: While I agree on that, I think it raises a larger question about i18n in a design system. Here we'd be using `ngx-translate`, some apps are using `transloco`, and this is, in the end, more of an app-dependant decision that shouldn't be handled here (like, how are we going to implement it in the design system at all? Does that mean that we'd have to provide json files for i18n and ask the consumer app to host them?).

There's a lot of different approaches we could discuss on that, like using an input for a11y label or providing a global i18n configuration object, or even providing an i18n translation approach from a configuration object given in `forRoot` or something.

packages/ng/callout/src/lib/callout/callout.component.html
jeremie-lucca: Also, it can be tiny AND small :
image

MarcFairbrother: Ok, `condensed` might have been a more appropriate naming choice but I don't mind if we stick to what already exists in the design system.

packages/ng/callout/src/lib/callout/callout.component.html
MarcFairbrother: Don't some ng-components already include translations that we use in apps with different i18n setups?

My point was more that the button needs a text alternative to the icon for accessibility purposes. This could be passed as an input as well if we don't want the component to include a translations file.

Supamiu: We have no translated values in lucca-front at the moment ^^.

Providing an alternative for the icon is for sure something we should provide, that's not the point that I'm trying to solve here.

My point is more that if we want proper a11y, we want this translated for sure, and having the library consumer provide the alt text doesn't really make sense as they shouldn't have to provide the alt text for something they didn't even add themselves (it'd mean having something like a `closeAltLabel` input that's used for the label).

Which is why you're raising a very interesting and complex problem that wed'll have to solve for the future too, by chosing which approach we want to use and stick to it for the new components (not to mention button itself will become a component, well a directive to be accurate, with its own alt input to make this easier).

MarcFairbrother: I agree, the alt text should be contained in the component.

There are some translated values here:

https://github.com/LuccaSA/lucca-front/blob/rc/packages/ng/establishment/src/lib/select/select-all/establishment-select-all.component.html

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2166/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/4569/ui-diff)

packages/ng/callout/src/lib/callout/callout.component.html
Supamiu: Oh, I see why I didn't find them, I was looking for `translate` or `transloco` usages in html files.

I'm not sure that's the best way to implement translation for the design system tbh, this one just uses a const exported from another file here: https://github.com/LuccaSA/lucca-front/blob/rc/packages/ng/establishment/src/lib/select/select-all/establishment-select-all.translate.ts which raises a maintenance issue.

I'm going to merge this PR regardless of that because the close mode isn't used in many places for now and I'll add a TODO to make sure it's not forgotten, but that topic is very interesting and will have to be solved for sure, we've got some ideas on how to make it as flexible as possible for everyone to use properly 😄

packages/ng/callout/src/lib/callout/callout.component.ts
```diff
@@ -0,0 +1,72 @@
...
+	 * Which palette should be used for the entire callout.
+	 * Defaults to none (inherits parent palette)
+	 */
+	palette: Palette = 'none';
+
+	@Input()
+	/**
+	 * Which size should the callout be? Defaults to medium
+	 */
+	size: SizeMS = 'm';
```
GuillaumeNury, line 31: I am a bit afraid of the future of this kind of type. We can have up to 9 sizes (XXXS, XXS, XS, S, M, L, XL, XXL, XXXL) but not all components can use all sizes. I would keep it simple with a `'m' | 's'` inlined here

stories/documentation/feedback/callout/callout-tiny.stories.ts
```diff
@@ -1,70 +1,52 @@
...
+import { CalloutComponent } from '@lucca-front/ng/callout';
+import { default as BasicStory } from './callout-basic.stories';
 
 export default {
-	title: 'Documentation/Feedback/Callout/Tiny',
+	...BasicStory,
+	title: 'Documentation/Feedback/Callout/Tiny'
+} as Meta;
+
+export const Template: StoryObj<CalloutComponent & { description: string }> = {
```
GuillaumeNury, line 10: I wonder if we still have to use separate files for stories. It would be easier to share some part if they were in the same file.

@jeremie-lucca does ZeroHeight support importing one story from a file containing multiple one?

packages/ng/callout/src/lib/callout/callout.component.html
GuillaumeNury: The current translation system is agnostic: it is the simplest way but it is not satisfying (all languages are bundled, we do not support variables nor plural, we dont have tools to ensure all translations exist in all languages).
Waiting for a translation refactor, you can stick with the current system:
- one file that declare a Record, an Interface and an InjectionToken ([example](https://github.com/LuccaSA/lucca-front/blob/63d46ba76e6433e07b672552395126dc848b88bc/packages/ng/modal/src/lib/modal.translate.ts))
- inject correct translations using `getIntl` ([example](https://github.com/LuccaSA/lucca-front/blob/6a13bbc422ea009679d54ff70e7ec5ae45100195/packages/ng/modal/src/lib/modal-panel.component.ts#L22))

packages/ng/callout/src/lib/callout/callout.component.scss
```diff
@@ -0,0 +1 @@
+@use '@lucca-front/scss/src/components/callout';
```
GuillaumeNury, line 1: I'm going to like to remove this line from our global styles.scss 🚀

packages/ng/callout/src/lib/callout/callout.component.ts
```diff
@@ -0,0 +1,72 @@
...
+	 * hidden. Hook on the `hidden` event emitter for that.
+	 */
+	removable = false;
+
+	@Input()
+	/**
+	 * Which icon should we display in the callout if any?
+	 * Defaults to no icon.
+	 */
+	icon: 'info' | 'success' | 'warning' | 'error' | 'help' | string;
```
GuillaumeNury, line 48: `'any' | 'string' | 'union' | string` is replaced with just `string`. While it is logical, it is sad because we loose auto-completion. Do you prefer to keep it like this for documentation purpose? Or move this list of icons in the comment?

GuillaumeNury (commented):
Thx Flavien for your contribution! Nicely commented code!
Do not hesitate to reach me about any of the comments I left in this PR.

packages/ng/callout/src/lib/index.ts
```diff
@@ -0,0 +1 @@
+export * from './callout/callout.component';
```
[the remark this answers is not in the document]
Supamiu: I moved everything to `packages/ng/callout`, as it'll be the same for the other ones too in a separate PR

stories/documentation/feedback/callout/callout-tiny.stories.ts
Supamiu: Because we won't have an answer until at least two weeks iirc, shouldn't we close this comment and tackle this in another future PR?

packages/ng/callout/src/lib/callout/callout.component.ts
Supamiu: As discussed together (but adding it here too for context), this is meant to be a helping hand using suggestions for devs, as the design team allows every icon to be used but recommends these ones. I'll create a helper type for it in a future PR.

GuillaumeNury (approved):

GuillaumeNury (approved):

LuccaIntegration:
:woman_cook: http://lucca-front.lucca.local/PR-2166/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/5399/ui-diff)

GuillaumeNury (approved):

packages/ng/core/src/lib/type/style.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Available CSS palettes
+ */
+export type Palette = 'success' | 'warning' | 'error' | 'primary' | 'grey' | 'none';
+
+/**
+ * Size that accepts only medium or small
+ */
+export type SizeMS = 'm' | 's';
```
GuillaumeNury, line 9: Reading the PR that moves all files, I saw this type that is not used anymore

Supamiu: Indeed, I'll make the change in #2210 though, to make sure it doesn't create new conflicts 😄

PR #2216, opened 2023-08-21
refactor(icons): new LuccaIcon type

Supamiu:
## Description

This PR adds the index.d.ts file to @lucca-front/icons so we can use it in ng package to type our `icon` inputs to `LuccaIcon`, which enables compilation time check and autocomplete inside your IDE to grab the icon
This file is generated from the selection.json file that icomoon gives us, then exported to be used in the ng package or in other consumers.

I also took this opportunity to refactor icons-related stories with the new CSF3 story format.

-----

BREAKING CHANGE
This removes the icons-list.js file, which, according to a github search, isn't used anywhere, but better safe than sorry so a check might be needed.
-----

PR #2302, opened 2023-10-02
feat: new lu-icon component

Supamiu:
## Description

This PR adds a new `lu-icon` component with autocompletion on the `icon` input.

You can use it like this: 

```html
<lu-icon icon="helpOutline"></lu-icon>
```
or even like this:
```html
<lu-icon icon="helpOutline"/>
```
-----

The component makes use of the new `LuccaIcon` type added in #2216 to have nice autocompletion and typechecking. It doesn't create a child span but instead adds the `lucca-icon icon-XXX` classes ot itself to avoid having to add one more node for nothing.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2302/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/7194/ui-diff)

packages/ng/icon/icon.component.ts
```diff
@@ -0,0 +1,29 @@
...
+	styleUrls: ['./icon.component.scss'],
+	changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class IconComponent implements OnChanges, OnInit {
+	#ngClass = inject(NgClass);
+
+	@Input({ required: true })
+	icon: LuccaIcon;
+
+	ngOnChanges(): void {
```
obraud, line 20: Ne faut-il pas préférer l'utilisation d'un setter sur les inputs?

Supamiu: Totalement, je l'ai mis en ngOnChanges en me disant que NgClass avait besoin d'être init avant, mais en le mettant en setter, ça fonctionne, j'aurais du tester cette solution aussi !

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2302/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/7196/ui-diff)

LucasPerrault (dismissed):

obraud (approved):

jeremie-lucca (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2302/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/7202/ui-diff)

packages/ng/icon/icon.component.ts
```diff
@@ -0,0 +1,30 @@
...
+
+	@HostBinding('attr.aria-hidden')
+	readonly ariaHidden = true;
+
+	@Input({ required: true })
+	set icon(icon: LuccaIcon) {
+		this.#ngClass.ngClass = {
+			[`icon-${icon}`]: true,
+		};
+	}
```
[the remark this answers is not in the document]
Supamiu: You're right, I didn't think of this kind of use case, I should just handle everything in ngClass.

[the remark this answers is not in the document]
Supamiu: Actually that's kind of a headache, because if we have two sources updating the same ngClass, it might not be possible to make them behave properly.

Imagine you have something like `<lu-icon icon="xxx" class="foo bar" ngClass="computedClasses$ | async"/>`, then we'd have an issue with two ngClass sources overrriding each other, and since ngClass doesn't have a way to merge these, I feel like this means that we can't really use ngClass in this case. @GuillaumeNury this is an interesting thing to consider regarding the `inject(ngClass)` approach that we might have overlooked 😄

packages/ng/icon/icon.component.html
```diff
@@ -0,0 +1,2 @@
+<span aria-hidden="true" class="lucca-icon icon-{{icon}} mod-{{size}} icon-color-{{color}}"></span>
+<span class="u-mask">{{alt}}</span>
```
vvalentin-lucca, line 2:
```suggestion
<span class="u-mask" *ngIf="alt">{{alt}}</span>
```

packages/ng/icon/icon.component.ts
```diff
@@ -0,0 +1,25 @@
+import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { LuccaIcon } from '@lucca-front/icons';
+
+@Component({
+	selector: 'lu-icon',
+	standalone: true,
+	imports: [CommonModule],
```
shajz, line 8: do we need it ? just NgIf if you accept @vvalentin-lucca's proposition

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2302/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/7331/ui-diff)

MarcFairbrother:
Are there any performance related risks when loading lots of Angular components on a web page? We could end up initialising dozens of components on some icon-heavy pages.

packages/ng/icon/icon.component.ts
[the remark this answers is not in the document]
GuillaumeNury: There is indeed a problem with this approach, I opened an [issue on Angular](https://github.com/angular/angular/issues/52072).

In the meantime, we can "copy" NgClass. But it is less elegant:

```ts
@Directive({
	selector: '[ngClazz]',
	standalone: true,
})
class NgClazz extends NgClass {}
```

packages/ng/icon/icon.component.html
Supamiu: I feel like this is pushing the dev into adding an alt but in the end, not using it if it's not here, meaning that in reality, it's optional. Idk, feels like lying on the API contract.

Supamiu:
> Are there any performance related risks when loading lots of Angular components on a web page? We could end up initialising dozens of components on some icon-heavy pages.

Angular is made for that so there's no performance issues to be worried of, especially with components like this one that are OnPush, standalone and require very little logic,

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2302/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/7394/ui-diff)

packages/ng/icon/icon.component.html
vvalentin-lucca: The same logic could be applied to the alt attribute on images: it's mandatory, but can be empty. The idea behind this is to say explicitly: "this is a decorative image, but I'm saying it explicitly with an empty alt and not implicitly with a missing alt."

vvalentin-lucca: (If you still think it's a nuance that can be overlooked, we can take the liberty of not making the parameter mandatory.)

packages/ng/icon/icon.component.ts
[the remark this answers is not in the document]
Supamiu: Well it no longer is a concern anyways because turns out we want an inner structure to be able to display an alt text

shajz (approved):

PR #2312, opened 2023-10-05
Add utilitities for *-block and *-inline properties

vvalentin-lucca:
## Description

Generate new utilities with logical properties. For example:
- `u-marginInline0`
- `u-paddingBlockXL`
- `u-borderBlockS`

-----

-----

### DIFF
--- a/packages/scss/src/commons/core.scss
+++ b/packages/scss/src/commons/core.scss
@@ -7,7 +7,7 @@
 
 $contents: '0';
 $boxModel: 'margin', 'padding', 'border';
-$boxDirection: '', 'top', 'bottom', 'left', 'right';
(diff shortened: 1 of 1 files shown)


martincadoux:
Trop bien ! Mais est-ce que ça ne vaudrait pas le coup d'ajouter également des valeurs `inline-start`, `inline-end`, `block-start` et `block-end` ?

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/7334/ui-diff)

vvalentin-lucca:
@martincadoux That could be a goal later on, but on this PR I'm mainly adding features that couldn't be achieved without using two classes until now.

In my opinion, functionality seems more urgent than naming.

MarcFairbrother (approved):

fbasmaison-lucca (approved):

PR #2334, opened 2023-10-12
fix icons nomenclature (open, close, bubble)

jeremie-lucca:
## Description

bubbles-answer > bubble-answer
bubbles-conversation > bubble-conversation
lock-closed > lock-close
lock-openned > lock-open

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2334/storybook

martincadoux:
If we want to evoke the state of the lock, I believe we should keep `icon-closed` as it is an adjective. The issue with the word « opened » was that it isn’t an adjective but the present perfect form of the verb « to open ».
You can think of the « Open / Closed » panels displayed on the front door of shops and restaurants as an example.

martincadoux (approved):

PR #2399, opened 2023-11-06
fix(simple-select): URI encoding done twice instead of just once

Supamiu:
## Description

This PR fixes a double encoding issue with apiVX directives for select components that turned `Lucca ` into `Lucca%25%20` due to it encoding once by hand (resulting in `Lucca%20`) and then having Angular encode params once again when assigned to HttpParams (resulting in `%` being encoded as `%25`).

Removing the manual encoding resulved the issue.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2399/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/9281/ui-diff)

g-drouard (approved):

PR #2401, opened 2023-11-07
refactor(wrappers): size now defaults to inherit rather than M

Supamiu:
## Description

We noticed an issue with icons, buttons and some other components that, when nested in wrappers, weren't behaving as they should. The issue being that since `size` defaulted to `M`, if the parent was using `mod-S` for its own size, the child component would not automatically be switched to `S` size as it already had a `mod-M` applied to it.

This created situations where icons are too large in a small component, for instance.

I fixed it by setting the default value for size to `undefined` (by just not defining it, not setting it explicitly) which won't apply any mod unless the size is explicitly passed as input value.

-----
-----

### DIFF
--- a/packages/ng/button/button.component.ts
+++ b/packages/ng/button/button.component.ts
@@ -17,7 +17,7 @@ export class ButtonComponent implements OnChanges {
 	#ngClazz = inject(NgClazz);
 
 	@Input()
-	size: 'M' | 'S' | 'XS' = 'M';
+	size: 'M' | 'S' | 'XS';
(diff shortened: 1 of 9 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2401/storybook

shajz (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/9307/ui-diff)

PR #2441, opened 2023-12-01
Colors variables

vvalentin-lucca:
## Description

In this PR I propose: 
- [X] rename the Sass color variables to something more in line with our final objectives;
- [X] better management of `!default` according to what is and isn't overloadable;
- [X] introduce final class and variable names (without deleting the old ones).

-----

You could therefore use, for example, `--palettes-neutral-700`, `--palettes-product-700`, or `--palettes-brand-700`.
Or overload scales locally for `spacings` or `border-radius`.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2441/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/10360/ui-diff)

packages/scss/src/commons/config.scss
```diff
@@ -96,120 +93,103 @@ $error: (
...
-	400: #acbbd7,
-	500: #96a6c5,
-	600: #667799,
-	700: #445473,
-	800: #2a3551,
-	900: #0b1732,
-) !default;
+);
 
-// Colors : Decoratives
```
MarcFairbrother, line 118: Shouldn't these colors be outside of the commons? I don't think that many softs are currently using them.

packages/scss/src/commons/base.scss
```diff
@@ -96,7 +96,7 @@
 
 		@each $palette in list.join(config.$palettes, config.$states) {
 			.palette-#{$palette} {
-				@each $shade in map.keys(config.$primary) {
+				@each $shade in map.keys(config.$product) {
```
MarcFairbrother, line 99: We could introduce a $shades object here instead of relying on one of the palettes.

vvalentin-lucca: And create an object just for this purpose?

MarcFairbrother: Yes if I understood the code we're juste using $primary or $product to loop over the existing shades and using those for each palette?

packages/scss/src/commons/config.scss
vvalentin-lucca: I've modified the structure a little so that you can configure LF like this and reduce the CSS variables: 

```
@use '@lucca-front/scss/src/commons/config' with (
	$newPalettes: [],
	$decoratives: [],
);
```

MarcFairbrother: Nice! Thanks

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/10377/ui-diff)

MarcFairbrother (dismissed):

fbasmaison-lucca (commented):
Super idée. Merci !

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2441/storybook

fbasmaison-lucca (approved):

PR #2445, opened 2023-12-04
fix inline message icon size + remove style encapsulation

jeremie-lucca:
## Description

- Fix icon size
- Remove style encapsulation in Angular component as discussed to avoid nested components imports issues 

-----

image

-----

### DIFF
--- a/packages/scss/src/components/inlineMessage/component.scss
+++ b/packages/scss/src/components/inlineMessage/component.scss
@@ -1,3 +1,5 @@
+@use '@lucca-front/icons/src/icon/exports' as icon;
+
 @mixin component {
   display: flex;
   gap: var(--components-inlineMessage-gap);
@@ -6,7 +8,7 @@
   color: var(--components-inlineMessage-color);
 
   .lucca-icon {
-    font-size: var(--components-inlineMessage-icon-fontSize);
+    @include icon.XS;
     color: var(--components-inlineMessage-icon-color);
     margin-top: 2px;
   }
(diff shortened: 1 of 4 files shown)


GuillaumeNury (dismissed):

packages/scss/src/components/inlineMessage/mods.scss
```diff
@@ -1,5 +1,10 @@
+@use '@lucca-front/icons/src/icon/exports' as icon;
+
 @mixin S {
 	--components-inlineMessage-fontSize: var(--sizes-XS-fontSize);
 	--components-inlineMessage-lineHeight: var(--sizes-XS-lineHeight);
-	--components-inlineMessage-icon-fontSize: .75rem;
+
+	.lucca-icon {
+		font-size: 0.75rem;
```
AlexiaAlarza, line 8: isn't it "`@include icon.XXS;`" ?

packages/scss/src/components/inlineMessage/component.scss
```diff
@@ -6,7 +8,7 @@
   color: var(--components-inlineMessage-color);
 
   .lucca-icon {
-    font-size: var(--components-inlineMessage-icon-fontSize);
+    @include icon.XS;
     color: var(--components-inlineMessage-icon-color);
     margin-top: 2px;
```
AlexiaAlarza, line 13: do we need to keep this margin instead of use `display flex` and `align-items center` on its parent ?

packages/scss/src/components/inlineMessage/mods.scss
jeremie-lucca: Ah yes, I forget this mixin size was added lately. Fixed.

packages/scss/src/components/inlineMessage/component.scss
jeremie-lucca: I checked with Clément and our icon must stay on top if the message is displayed on multiple line, so we can't center it :'(

AlexiaAlarza (approved):

GuillaumeNury (approved):

PR #2562, opened 2024-02-13
Tabindex focus-visible

vvalentin-lucca:
## Description

A default style for focusable elements via a tabindex. (This is particularly useful for tooltips.)

-----

The `:where` keeps specificity of the selector at 0.

-----

### DIFF
--- a/packages/scss/src/commons/base.scss
+++ b/packages/scss/src/commons/base.scss
@@ -94,6 +94,10 @@
 			font-weight: 600;
 		}
 
+		*:where([tabindex='0']:focus-visible) {
+			@include a11y.focusVisible($borderRadius: var(--commons-borderRadius-M));
+		}
+
 		@each $palette in list.join(config.$palettes, config.$states) {
 			$paletteExists: map.get(config.$palettesInterpolation, $palette);
 


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2562/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/13164/ui-diff)

packages/scss/src/commons/base.scss
```diff
@@ -94,6 +94,10 @@
 			font-weight: 600;
 		}
 
+		*:where([tabindex='0']:focus-visible) {
+			@include a11y.focusVisible($borderRadius: var(--commons-borderRadius-M));
+		}
```
MarcFairbrother, line 99: Not sure about this, there are various instances where I want the focus to display around a parent element rather than the focusable element itself for visual reasons. I would have to override this rule to avoid a double focus effect.

vvalentin-lucca: You're right, I'll close until I find something better.

MarcFairbrother: Thanks!

PR #2596, opened 2024-02-27
chore: fix style for proper sizing in dialogs

Supamiu:
## Description

-----

-----

### DIFF
--- a/packages/ng/dialog/dialog.service.ts
+++ b/packages/ng/dialog/dialog.service.ts
@@ -1,4 +1,4 @@
-import { inject, Injectable, Injector, RendererFactory2 } from '@angular/core';
+import { inject, Injectable, Injector, Renderer2 } from '@angular/core';
 import { LuDialogConfig, LuDialogRef, LuDialogResult } from './model';
 import { Dialog, DialogRef } from '@angular/cdk/dialog';
 import { isObservable, merge, of, take } from 'rxjs';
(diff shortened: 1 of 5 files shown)


packages/ng/dialog/dialog.service.ts
```diff
@@ -50,6 +52,10 @@ export class LuDialogService {
 			...(config.cdkConfigOverride || {}),
 		});
 
+		if (cdkRef.componentRef) {
+			this.#renderer.setStyle(cdkRef.componentRef.location.nativeElement, 'display', 'contents');
+		}
```
GuillaumeNury, line 57: Can't we get a renderer using `cdkRef.componentRef.injector.get(Renderer2)` ?

Supamiu: We can't because we're in a service so we don't have access to `Renderer2`

Supamiu: Oh wait, I just realized you're right indeed 😄 I read that wrong and thought you meant to just use inject.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2596/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/13862/ui-diff)

jeremie-lucca (approved):

GuillaumeNury (approved):

PR #2623, opened 2024-03-04
fix(select): position not updating when new options are set

Supamiu:
## Description

Fixed a bug that happened with options were set after the panel is opened, which happens quite often with `*apiV4`.
When options are set, if the panel is too far on the right hand side of the screen and the data requires the panel to be larger than the select input's width, then it would expand to the right no matter what, at least when opening it for the first time.

This was due to the panel's position not being updated after new options are set, which is what this PR fixes.

-----
-----

### DIFF
--- a/packages/ng/core-select/input/select-input.component.ts
+++ b/packages/ng/core-select/input/select-input.component.ts
@@ -1,12 +1,14 @@
 /* eslint-disable @angular-eslint/no-output-on-prefix */
 import { OverlayConfig, OverlayContainer } from '@angular/cdk/overlay';
 import {
+	booleanAttribute,
 	ChangeDetectorRef,
 	Directive,
 	ElementRef,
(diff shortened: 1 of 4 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2623/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/14050/ui-diff)

GuillaumeNury (dismissed):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2623/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/14056/ui-diff)

jeremie-lucca (approved):

PR #2664, opened 2024-03-19
feat(form-fields): new `lu-number-input` and various input improvements

Supamiu:
## Description

- add `statusControl` input to `lu-form-field` to be able to have a parent control drive a child's validity display.
This means that when adding a `statusControl`, field's `invalid` flag (aka `is-error` class) logic becomes `this.statusControl.invalid || this.control.invalid`. Which is useful when a parent validator controls invalidity (for instance when two values have to match each other or follow a common rule).

- add new `lu-number-input` to handle `inpout[type=number]` in a betetr fashion.
Because we have component inputs that are used only for this one, we figured having a separate component would make sense.
Basically, `lu-number-input` is a `lu-text-input` without the ability to show password toggle and with added inputs: `min`, `max`, `step` and `noSpinButtons`.

- add new `errorInlineMessage` input to `lu-form-field` to be able to provide a message to show only if the control is `invalid` (or, if you have `statusControl` provided, `this.statusControl.invalid || this.control.invalid`. 

## ⚠️BREAKING CHANGES⚠️

This PR breaks `noSpinButtons` and `step` inputs on `lu-text-input` added in 1eb571b391bbb4f1bcf980d779964c47dad40c98 because it removes them in favor of `lu-number-input` now using them (as they were only for number inputs).

It also removes `number` from the list of valid values for `lu-text-input`'s `type` input, as you should now be using `lu-number-input` instead.

-----

This PR includes unmerged changed from `master` that weren't sent to `rc` when the branch was created, most of them should disappear once a new `master` to `rc` sync will be done but not the merge commit, which can be removed if needed but considering we'll squash, I'm not sure that's needed.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2664/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/14832/ui-diff)

packages/ng/form-field/form-field.component.ts
```diff
@@ -165,7 +168,7 @@ export class FormFieldComponent implements OnChanges, OnDestroy, DoCheck {
 		if (this.#control) {
 			// invalid management
 			const previousInvalid = this.invalid;
-			this.invalid = this.#control.invalid && this.#control.touched;
+			this.invalid = (this.#control.invalid || this.statusControl?.invalid) && this.#control.touched;
```
thomasmarcelin, line 177: If `control` is `invalid` and `statusControl` is `valid`, the `FormField` will still be invalid and the `statusControl` won't drive the validity of the field. Is this the expected result ?

Supamiu: Maybe it means that we should rename it but yeah, because you can have validators on both a parent and its child, and one of them being invalid means that you want to mark child as invalid regardless. That's the use case here.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2664/storybook

thomasmarcelin (dismissed):

GuillaumeNury (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/15153/ui-diff)

jeremie-lucca:
@Supamiu Could we remove type option on `lu-number-input` as this component is 100% dedicated to numbers? (I think we can only to keep it on `lu-textfield-input`). 
Everything else seems perfect 🙌

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2664/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/15237/ui-diff)

jeremie-lucca (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2664/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/15241/ui-diff)

PR #2673, opened 2024-03-21
fix(select): panel position not updating on api data changes 

Supamiu:
## Description

This was caused by the fact that we directly connected data source to `options$` instead of using the setter, which is also updating panel position.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2673/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/14961/ui-diff)

lamstutz (approved):

jeremie-lucca (approved):

PR #2690, opened 2024-03-29
fix(schematics): missing required schema.json

shajz:
## Description

The current generators are not working with an nx workspace :

Screenshot 2024-03-29 at 15 42 48

This is because `nx` is looking for the mandatory schema.json file that's supposed to be part of the schematic (https://angular.io/guide/schematics-for-libraries#configure-the-new-schematic)

This can be seen here : https://github.com/nrwl/nx/blob/ef81455b6492fee266c0d6ddfc41a79dc2041f8b/packages/nx/src/command-line/generate/generator-utils.ts#L41-L45

Schematics without parameters can simply be declared with an empty object (as is the case here)

-----

### DIFF
--- a/packages/ng/schematics/collection.json
+++ b/packages/ng/schematics/collection.json
@@ -3,11 +3,13 @@
 	"schematics": {
 		"ng-add": {
 			"description": "Add my lucca-front to the project.",
-			"factory": "./ng-add/index#ngAdd"
+			"factory": "./ng-add/index#ngAdd",
+			"schema": "./ng-add/schema.json"
 		},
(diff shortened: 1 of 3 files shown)


shajz:
cc @guillerot @Supamiu @GuillaumeNury

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2690/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/15373/ui-diff)

PR #2694, opened 2024-04-02
Fix icon size fallback

vvalentin-lucca:
## Description

The `--icon-size` variable remains undefined so that it can be overloaded, so we use another variable name for the fallback.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2694/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/15495/ui-diff)

PR #2722, opened 2024-04-11
fix(empty-state): http request for icon still fired even when it's null

Supamiu:
## Description

Empty state components have an optional icon but when it's not provided, the request is still fired, creating an error.

This is fixed by only showing icon when it's here and handling null gracefully in safe svg pipe to handle possible async pipe use cases.

-----

-----

### DIFF
--- a/packages/ng/empty-state/empty-state-page/empty-state-page.component.html
+++ b/packages/ng/empty-state/empty-state-page/empty-state-page.component.html
@@ -8,9 +8,9 @@
 >
 	<div class="emptyState-container">
 		<div class="emptyState-content">
-			<div class="emptyState-content-icon" aria-hidden="true" [innerHtml]="icon | luSafeExternalSvg"></div>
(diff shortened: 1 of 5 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2722/storybook

shajz (approved):

shajz:
Shouldn't this fix point to master?

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/15917/ui-diff)

Supamiu:
Pointing to rc because it's not critical, just throwing a non-blocking error, showing in console and 17.4 is right around the corner.

GuillaumeNury (approved):

GuillaumeNury:
@Supamiu isn't the 17.4 PR already merged into master ?

PR #2731, opened 2024-04-12
fix(user-popover): fix initial position after employee is loaded

Supamiu:
## Description

Fix the initial position when opening user-popover for the first time

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2731/storybook

jeremie-lucca (approved):

PR #2742, opened 2024-04-17
Less user picture API calls using lazy loading attribute

GuillaumeNury:
## Description

Less user picture API calls using lazy loading attribute.

-----

### DIFF
--- a/packages/ng/user/picture/user-picture.component.html
+++ b/packages/ng/user/picture/user-picture.component.html
@@ -1,5 +1,5 @@
 <ng-container *ngIf="hasPicture; else displayInitials">
-	<img class="picture" [src]="pictureHref" alt="" (error)="pictureError()" />
+	<img class="picture" [src]="pictureHref" alt="" (error)="pictureError()" [loading]="imageLoadingAttribute" />
 </ng-container>
 
 <ng-template #displayInitials>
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2742/storybook

Supamiu (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16129/ui-diff)

packages/ng/user/picture/user-picture.component.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="hasPicture; else displayInitials">
-	<img class="picture" [src]="pictureHref" alt="" (error)="pictureError()" />
+	<img class="picture" [src]="pictureHref" alt="" (error)="pictureError()" [loading]="imageLoadingAttribute" />
```
MarcFairbrother, line 2: We were discussing the `alt` attribute with @vvalentin-lucca and @ymalie-lucca 

Could we pass the employee's fullName by default and have the options to have either an empty alt or a custom string?

GuillaumeNury: Last time we discussed about it, we wanted to avoid a duplication of the name. The user-picture component is often used along user name. A screen reader would read "Daniel Hernandez Daniel Hernandez".
If the `alt` attribute have to be customizable, we can do it, but I would keep the default value as empty.
Anyway, it is not the purpose of this PR

PR #2749, opened 2024-04-22
Add wrapper with .u-mask

vvalentin-lucca:
## Description

I think this will be safer to avoid any side effects during tests.

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16259/ui-diff)

PR #2750, opened 2024-04-22
refactor(lu-tooltip): simplify code and implement ellipsis system

Supamiu:
## Description

Refactor of the `luTooltip` directive because it was relying on a lot of common code with `LuPopover` code while it's not the same thing at all and doesn't require the same approaches to various things (like position). It was also not acting like we want it to act regarding ellipsis (ellipsis tooltip should be hidden from screen readers and detection needs to be accurate).

ℹ️ This is a non-breaking refactor, everything will work as expected with existing code, with better accessibility and better ellipsis detection

The refactor includes:

- Direct use of `@angular/cdk` for the tooltip logic instead of using the one hidden in `ALuPopoverTrigger` for better granularity on features we want/don't want
- Removal of inheritence for easier maintenance, in order to keep the code simple to read
- `luTooltipWhenEllipsis` is now accurate in all the cases tested inside our testing story ([stories/documentation/overlays/tooltip/tooltip-ellipsis.stories.ts](https://github.com/LuccaSA/lucca-front/compare/rc...refactor/tooltips#diff-165d3c8d9be8269d4e08ed7621a9e4eadc171a71bca3f308e5fa651b6490f3d4))
- When a tooltip is triggered by ellipsis, it's not seen by Screen Readers to avoid double reading of the same content
- `luTooltipWhenEllipsis` can now be used with an empty `luTooltip`, making the tooltip automatically getting its content from `host.nativeElement.innerText`, thus removing any HTML present inside of it to only keep rendered text as string.

-----

-----


packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
```diff
@@ -1,156 +1,339 @@
...
-	@Input('luTooltipWhenEllipsis') public set inputWhenEllipsis(we: boolean) {
-		this.whenEllipsis = we;
-	}
+	@Input({ transform: booleanAttribute })
+	luTooltipWhenEllipsis = false;
+
+	resize$ = new Observable((observer) => {
+		new ResizeObserver(() => {
+			observer.next();
+		}).observe(this.#host.nativeElement);
```
GuillaumeNury, line 64: We could "unobserve" in the teardown logic of this observable: 

```ts
new Observable(observer => {
	// Observe here

	// Teardown logic
	return () => {
		// Unobserve here
	}
})

```

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
```diff
@@ -1,156 +1,339 @@
...
-	// private set tabindex(i: number = null) {
 
-	// }
+	#generatedId = `${this.#host.nativeElement.tagName.toLowerCase()}-tooltip-${nextId++}`;
 
-	/** accessibility attribute - dont override */
-	@HostBinding('attr.id') get _attrId() {
-		return this._triggerId;
+	@HostBinding('attr.id')
+	get _attrId() {
```
GuillaumeNury, line 94: No need for a getter here?

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
```diff
@@ -1,156 +1,339 @@
...
+	 * you'd get true while it should be false and vice-versa, because of rounding.
+	 *
+	 * We could also use getBoundingClientRect() but it only considers text content, meaning that if ellipsis is caused by
+	 * any margin or padding, it won't be detected
+	 *
+	 * @private
+	 */
+	private hasEllipsis(): boolean {
+		if (window.getComputedStyle(this.#host.nativeElement).textOverflow !== 'ellipsis') {
+			return false;
```
GuillaumeNury, line 212: should we log a warning if the user ask for `whenEllispsis` but `text-overflow` is not correct?

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
```diff
@@ -1,156 +1,339 @@
...
+			]);
+	}
+
+	private invertVerticalPos(y: VerticalConnectionPos) {
+		if (y === 'top') {
+			y = 'bottom';
+		} else if (y === 'bottom') {
+			y = 'top';
+		}
+		return y;
```
jbosi, line 327: Plutôt que de muter `y` puis de le renvoyer, on ne peut pas plus simplement renvoyer une `VerticalConnectionPos` ?

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
```diff
@@ -1,156 +1,339 @@
...
-		this._panel = componentRef.instance;
-		this._panel.content = this._tooltipContent;
+	/**
+	 * Hacky af but let's explain everything
+	 * This method checks for ellipsis by cloning the node and checking its width against original element.
+	 *
+	 * We used to do this using scrollWidth but the thing is, it's a rounded value. Sometimes,
+	 * you'd get true while it should be false and vice-versa, because of rounding.
+	 *
+	 * We could also use getBoundingClientRect() but it only considers text content, meaning that if ellipsis is caused by
```
GuillaumeNury, line 205: Paddings are included in `getBoundingClientRect` as well as borders. I don't get why we should clone the element. `boundingRect` on host element should be enough?

GuillaumeNury (commented):
🚀

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16284/ui-diff)

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
Supamiu: Totalement si, c'est du code que je n'ai pas assez simplifié par rapport à l'original (l'idée étant de garder les mêmes inputs de positionnement donc autant garder le code pour assurer une full compat' avec l'original).

Je vais changer ça.

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
Supamiu: I'm not sure I understand the goal you want to achieve with this?

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
Supamiu: We could but that would possibly cause a lot of warnings in various places since this is not a new component, it's a refactor of an existing one.

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
Supamiu: There's a more detailed answer than I could provide here: https://stackoverflow.com/a/64747288/4102561

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2750/storybook

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
Supamiu: Nvm, got it.

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16339/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2750/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16343/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2750/storybook

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2750/storybook

vvalentin-lucca (approved):

PR #2761, opened 2024-04-25
[Secondary nav] Fix inverted deprecated  alert color when item is active

jeremie-lucca:
## Description

Fix inverted deprecated  alert color when item is active

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2761/storybook

PR #2768, opened 2024-04-29
[Datepicker] Invalid cover

jeremie-lucca:
## Description

- Fix placeholder color with `is-invalid` class (previously only `is-error`)
- Date icon should stay grey in error/invalid state to fit with our Design System

-----

image

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16623/ui-diff)

PR #2776, opened 2024-05-06
fix(form-field): get validators from descendants to handle deep inputs

Supamiu:
## Description

Because we were only querying the first layer of content children, we were only fetching controls from it and not descendants, meaning that in cases like this:

```html
<lu-form-field>
  <div class="inputWrapper">
    <lu-text-input type="text"></lu-text-input>
    <button luButton="text" type="button"></button>
  </div>
</lu-form-field>
```

The `lu-text-input` component wasn't behaving properly with `aria-*` flags and the field wasn't detected as required.

-----

-----

### DIFF
--- a/packages/ng/form-field/form-field.component.ts
+++ b/packages/ng/form-field/form-field.component.ts
@@ -43,12 +43,12 @@ export class FormFieldComponent implements OnChanges, OnDestroy, DoCheck {
 
 	#doCheck$ = new Subject<void>();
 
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2776/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16832/ui-diff)

GuillaumeNury (approved):

PR #2779, opened 2024-05-07
fix(callout): default callout size

shajz:
## Description

Callout says it defaults to `'M'` but it doesn't.

-----

### DIFF
--- a/packages/ng/callout/callout/callout.component.ts
+++ b/packages/ng/callout/callout/callout.component.ts
@@ -32,7 +32,7 @@ export class CalloutComponent {
 	/**
 	 * Which size should the callout be? Defaults to medium
 	 */
-	size: 'M' | 'S';
+	size: 'M' | 'S' = 'M';
 
 	@Input({ transform: booleanAttribute })
 	/**


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2779/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16918/ui-diff)

Supamiu:
While the size doesn't explicitly default to `'M'`, it technically does, and shouldn't be enforced explicitly.

Imagine something like a larger LF component which also has sizing, using `lu-callout` internally, a quick structure could be:

```html
<div class="something-component">
...content
<lu-callout></lu-callout>
</div>
```
And in SCSS:
```scss
.something-component {
 .callout {
    @include S;
 }
}
```

Having an explicit default size would render this impossible because having a `size = 'M'` means we'd have the `mod-M` class on the callout which would force the size dans make SCSS size manipulation impossible.

Ping @vvalentin-lucca for more details if needed.

vvalentin-lucca:
By default, the size of components is set to M, but this is not explicit, allowing you to inherit a size set higher by an enclosing component.

- So if a button (default size M) is in a callout of size S, it will automatically change to size S.
- Inversely, a button explicitly set to size M, even in a callout of size S, will remain in size M.

Supamiu:
We should indeed use `[ngClass]` to avoid having `mod-undefined`

jeremie-lucca:
The way mod classes are managed in Angular components has been improved lately here: https://github.com/LuccaSA/lucca-front/pull/2984

And indeed, `mod-M` shouldn't be applied by default on components, as it would increase the specificity, leading to break nested components, as described here: https://github.com/LuccaSA/lucca-front/pull/2779#issuecomment-2107011964 

I'm closing the issue ✌️

PR #2782, opened 2024-05-13
fix(radio): disabled attribute and autogenerated name

Supamiu:
## Description

- `<lu-radio disabled="false">...` is now properly enabled, as we're using the native `disabled` attribute, which evaluates `false` as `true` because it's non-null.
- `name` attribute is now set by the parent group to allow multiple `lu-radio-roup-input` to behave properly in the same page.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2782/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16951/ui-diff)

g-drouard (approved):

GuillaumeNury (approved):

PR #2788, opened 2024-05-13
Add support for the `rows` attribute

### LINKED ISSUE #2752: [Textarea] Add support for the `rows` attribute
The **textarea** Angular component currently has a fixed height equivalent to three rows of text.

Would it be possible to let us add the `rows` attribute (via a `[rows]` input for instance), and let the textarea natively size itself based on its value instead of a fixed height?

Alternatively, we could use a custom property instead of a fixed value to set the height of the textarea:

```
--components-textfield-rows: 3;

  .textField-input-value:is(textarea) {
    height: calc(var(--components-textfield-rows) * 1lh + var(--component-textField-padding) * 2);
  }
```

vvalentin-lucca:
## Description

Fix #2752 

-----

-----

### DIFF
--- a/packages/ng/forms/textarea-input/textarea-input.component.ts
+++ b/packages/ng/forms/textarea-input/textarea-input.component.ts
@@ -1,6 +1,6 @@
 import { ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@angular/core';
-import { InputDirective } from '@lucca-front/ng/form-field';
 import { ReactiveFormsModule } from '@angular/forms';
+import { InputDirective } from '@lucca-front/ng/form-field';
 import { injectNgControl } from '../inject-ng-control';
 import { NoopValueAccessorDirective } from '../noop-value-accessor.directive';
 
@@ -18,4 +18,7 @@ export class TextareaInputComponent {
 
 	@Input()
 	placeholder: string = '';
+
+	@Input()
+	rows?: number;
 }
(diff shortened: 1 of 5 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/16975/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2788/storybook

jeremie-lucca (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17006/ui-diff)

packages/ng/forms/textarea-input/textarea-input.component.ts
```diff
@@ -18,4 +18,7 @@ export class TextareaInputComponent {
 
 	@Input()
 	placeholder: string = '';
+
+	@Input()
+	rows?: number;
```
shajz, line 23: Should we set 2 as the default value? It's the official default value.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#rows
```suggestion
	rows: number = 2;
```

vvalentin-lucca: Good point. 

The idea was to respect the design, but the question also existed of modifying it.

vvalentin-lucca: Good point.
The idea was to respect the design, but the question also existed of modifying it.

Does this make the parameter mandatory? I think not.

shajz: No, if the parameter is not provided, the default value will be used.

vvalentin-lucca: On the design side, the intention is to stay with 3 lines by default, so we'll keep it that way.

Supamiu (approved):

PR #2794, opened 2024-05-15
fix(core-select): wrong naming for if-clue-empty startegy

GuillaumeNury:
## Description

* The condition was the opposite of what the old naming said
* Added a new strategy with the previous behavior

-----

Even if it is a breaking change, nobody uses it at the moment.

-----


shajz (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2794/storybook

Supamiu (approved):

PR #2799, opened 2024-05-17
[Toast] Fix misaligned kill button in Safari iOS

belabb-z:
## Description

Toast kill button was misaligned only in iOS browsers

-----

iOS sets a default padding on this button, which is not there in other browsers ; unsetting it fixes the issue

Before:
image

After:
image

-----


shajz (approved):

jeremie-lucca (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17171/ui-diff)

PR #2804, opened 2024-05-21
fix(select): strange selection behavior on empty search result

GuillaumeNury:
## Description

Pressing "Enter" when making a search without result was selecting either `null` or the last active item. This PR resets the active item when no options are displayed and stop sending `null` as a value.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2804/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17275/ui-diff)

Supamiu (approved):

PR #2814, opened 2024-05-23
Empty state titles

### LINKED ISSUE #2780: [EmptyState] Title font-size should not depend on it's semantic level
`EmptyState` title font-size (and line-height) is currently tied to it's semantic level (h1, h2, h3...).

It should not be the case : the `font-size` should always be `--sizes-L-fontSize` or `var(--sizes-XXL-fontSize)` and carried by the class `emptyState-content-heading`, no matter what semantic level the title realy is in the DOM.

It also seems that the Angular component for the `emptyState` always use `h1` or `h3` for it's title level and do not provide an option to change this. In many cases it won't be the correct semantic title level to use.

vvalentin-lucca:
## Description

Fix #2780 

- [x] The style of empty state titles no longer depends on their semantic level.
- [x] A parameter is added to Angular components to make this title level configurable.

-----

-----

Capture d’écran 2024-05-23 à 17 11 57

### DIFF
--- a/packages/ng/empty-state/empty-state-page/empty-state-page.component.html
+++ b/packages/ng/empty-state/empty-state-page/empty-state-page.component.html
@@ -10,7 +10,7 @@
 		<div class="emptyState-content">
 			<div *ngIf="icon" class="emptyState-content-icon" aria-hidden="true" [innerHtml]="icon | luSafeExternalSvg"></div>
 			<div class="emptyState-content-text">
-				<h1 class="emptyState-content-heading">{{ title }}</h1>
+				<div role="heading" [attr.aria-level]="hx" class="emptyState-content-heading">{{ title }}</div>
 				<p>
 					<ng-container *luPortal="description"></ng-container>
 				</p>
--- a/packages/ng/empty-state/empty-state-section/empty-state-section.component.ts
+++ b/packages/ng/empty-state/empty-state-section/empty-state-section.component.ts
@@ -1,7 +1,7 @@
+import { NgIf } from '@angular/common';
 import { booleanAttribute, ChangeDetectionStrategy, Component, Input, ViewEncapsulation } from '@angular/core';
 import { Palette, PortalContent, PortalDirective } from '@lucca-front/ng/core';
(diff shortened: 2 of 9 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17341/ui-diff)

packages/ng/empty-state/empty-state-section/empty-state-section.component.ts
```diff
@@ -36,4 +36,7 @@ export class EmptyStateSectionComponent {
 		required: true,
 	})
 	description: PortalContent;
+
+	@Input()
+	hx: number = 3;
```
Supamiu, line 41:
```suggestion
	hx: 1 | 2 | 3 | 4 | 5 | 6 = 3;
```
If we want to be exhaustive regarding the possible values, this union type would be better (and include autocompletion for possible values)

Supamiu (dismissed):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2814/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17384/ui-diff)

packages/ng/empty-state/empty-state-page/empty-state-page.component.html
```diff
@@ -10,7 +10,7 @@
 		<div class="emptyState-content">
 			<div *ngIf="icon" class="emptyState-content-icon" aria-hidden="true" [innerHtml]="icon | luSafeExternalSvg"></div>
 			<div class="emptyState-content-text">
-				<h1 class="emptyState-content-heading">{{ title }}</h1>
+				<div role="heading" [attr.aria-level]="hx" class="emptyState-content-heading">{{ title }}</div>
```
fbasmaison-lucca, line 13: Why not using `<h{{ hx }}>…</h{{ hx }}>`?

shajz: I don't think that's possible with Angular. It would need to be a custom component that uses `*ngIf` directives to pick the right element.

jeremie-lucca (approved):

PR #2816, opened 2024-05-23
fix(core-select): do not send empty fields / wrong operation parameter name

GuillaumeNury:
## Description

* do not send empty fields
* wrong operation parameter name

-----

### DIFF
--- a/packages/ng/core-select/user/users.directive.ts
+++ b/packages/ng/core-select/user/users.directive.ts
@@ -130,7 +130,7 @@ export class LuCoreSelectUsersDirective<T extends LuCoreSelectUser = LuCoreSelec
 		const params = {
 			fields: this.#userFields,
 			...(this._filters() ?? {}),
-			...(this._operationIds() ? { operationIds: this._operationIds().join(',') } : {}),
+			...(this._operationIds() ? { operations: this._operationIds().join(',') } : {}),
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2816/storybook

packages/ng/core-select/user/users.directive.ts
```diff
@@ -130,7 +130,7 @@ export class LuCoreSelectUsersDirective<T extends LuCoreSelectUser = LuCoreSelec
 		const params = {
 			fields: this.#userFields,
 			...(this._filters() ?? {}),
-			...(this._operationIds() ? { operationIds: this._operationIds().join(',') } : {}),
+			...(this._operationIds() ? { operations: this._operationIds().join(',') } : {}),
```
shajz, line 133: wasn't this fixed somewhere already? I'm having "déjà vu" :D

shajz (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17344/ui-diff)

packages/ng/core-select/user/users.directive.ts
GuillaumeNury: it was fixed a few lines above, for regular options. The present fix is for the "Me: " option 🙈

PR #2818, opened 2024-05-28
fix(tooltip): wrong disable transform

GuillaumeNury:
## Description

We cannot disable the new tooltip because of the wrong input transformer.

-----

-----


bastienguillon (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2818/storybook

PR #2838, opened 2024-06-04
Show index table link

vvalentin-lucca:
## Description

Users have expressed the need to be able to open links in tables to new tabs  (via the right-click menu, for example).
To do this, these links must not be hidden and must be visible in the interface.

`.indexTable-body-row-cell-action` is therefore deprecated and replaced by `.indexTable-body-row-cell-link`, which is its unmasked equivalent, and following @BertrandPodevin's idea, we extend this link so that it can be operated on the whole line, without requiring an `overflow: hidden`, `z-index`, or superimposing itself on other interactive elements.

-----

-----

Capture d’écran 2024-06-04 à 12 04 23

### DIFF
--- a/packages/scss/src/components/indexTable/component.scss
+++ b/packages/scss/src/components/indexTable/component.scss
@@ -12,6 +12,7 @@
 	border-radius: var(--components-indexTable-border-radius);
 	background-color: var(--components-indexTable-background-color);
 	color: var(--palettes-neutral-800);
+	contain: paint;
 
 	@at-root ($atRoot) {
 		.indexTable-head {
@@ -39,7 +40,6 @@
 
 		.indexTable-body-row {
 			border-radius: var(--commons-borderRadius-L);
-			cursor: pointer;
 		}
 
 		.indexTable-head-row-transparentCell,
@@ -56,6 +56,9 @@
 		.indexTable-body-row-transparentCell {
 			vertical-align: middle;
 			padding: var(--pr-t-spacings-50) 0;
+			position: relative;
+			z-index: 3;
+
 			&:not(:first-child) {
 				padding-right: var(--components-indexTable-cell-padding);
 			}
@@ -80,6 +83,10 @@
 
 			&:last-child {
 				--components-indexTable-cell-padding-right: var(--components-indexTable-cell-padding);
+			}
+
+			&:not(:has(button, a, input, textarea, select, [tabindex='0'])) {
+				pointer-events: none;
 			}
 		}
 
@@ -183,6 +190,12 @@
 
 		.indexTable-body-row-cell-link {
 			text-decoration: none;
+
+			&::after {
+				content: '';
+				position: absolute;
+				inset: 0 -9999px;
+			}
 
 			&,
 			&:hover {
--- a/packages/scss/src/components/indexTable/component.scss
+++ b/packages/scss/src/components/indexTable/component.scss
@@ -176,9 +176,22 @@
 			margin: 2px 0 0 var(--pr-t-spacings-50);
 		}
 
-		// Hidden element that take focus when the row is focused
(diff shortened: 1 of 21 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2838/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/17890/ui-diff)

BertrandPodevin:
I find this change somewhat weird because 
- the right click will only work if the user interact exatly on the column containing the link but, without any visual indicator he has no way to know wich one it is or even that he can do it. 
- for screen reader, the link content won't make sense. ex `<a>Daniel Hernandez</a>` or `<a>CP 2023 / 2024</a>`

I think that for it to realy work, i think that we should embed the content of every column into a link and find a way to provide some context to screen reader, like "show details", while also prevent redundancy and tab hell.

vvalentin-lucca:
@BertrandPodevin I agree with you.

We think it's a better solution than the completely hidden version of the link, but it's still not perfect. We thought of putting several links in each column, with a `tabindex="-1"` and a `role="presentation"`, but it makes the code very heavy.

BertrandPodevin:
Maybe a `role="link"` directly on the `tr` + some javascript magics could do the trick ?

https://developer.mozilla.org/fr/docs/Web/Accessibility/ARIA/Roles/link_role

vvalentin-lucca:
No, it overwrites the `row` role of the `tr` and the table row disappears from the restitution.

fbasmaison-lucca:
Even though lines are supposed to be clickable since we are on _Index Tables_, some are not yet implemented with a link and moving to a `cursor: auto` can actually reduce the feeling of being deceived by the interface, `cursor: pointer` being [forced at the moment](https://github.com/LuccaSA/lucca-front/blob/8cca3edee35e1f12b9a824ca15676cd2704fc66d/packages/scss/src/components/indexTable/component.scss#L42).

Would changing the pointer to only the parts that are really clickable help for discovery?

Clicking anywhere on the row could still open as expected by us, but the UI feedback would actually reflect the semantic of the elements being hovered.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2838/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/18014/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2838/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/18038/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2838/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/20899/ui-diff)

packages/scss/src/components/indexTable/component.scss
```diff
@@ -81,6 +84,10 @@
 			&:last-child {
 				--components-indexTable-cell-padding-right: var(--components-indexTable-cell-padding);
 			}
+
+			&:not(:has(button, a, input, textarea, select, [tabindex='0'])) {
```
BertrandPodevin, line 88: Maybe add `details` ? Super edge case but our callout disclosure use it

BertrandPodevin:
Maybe add a `mod-allowTextSelection` that disable the whole thing ?

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2838/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/20912/ui-diff)

BertrandPodevin (approved):

PR #2956, opened 2024-07-16
flex-start/end > start/end

vvalentin-lucca:
## Description

The new syntax is now well supported everywhere and more consistent between grid and flex.

-----

-----

### DIFF
--- a/packages/scss/src/commons/core.scss
+++ b/packages/scss/src/commons/core.scss
@@ -13,8 +13,10 @@ $gaps: 'column-gap', 'row-gap', 'gap';
 $displays: 'block', 'flex', 'inline', 'inline-flex', 'inline-block', 'contents', 'none';
 $direction: 'column', 'column-reverse', 'row', 'row-reverse';
 $wrap: 'wrap', 'nowrap', 'wrap-reverse';
-$justify: 'flex-start', 'flex-end', 'space-around', 'space-between', 'space-evenly', 'center';
-$align: 'flex-start', 'flex-end', 'baseline', 'stretch', 'center';
+// flex-start and flex-end are deprecated
+$justify: 'flex-start', 'start', 'flex-end', 'end', 'space-around', 'space-between', 'space-evenly', 'center';
+// flex-start and flex-end are deprecated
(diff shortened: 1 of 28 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-2956/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/19693/ui-diff)

fbasmaison-lucca:
Did we manually test all implementations are correct? Because I’m not entirely sure a search and replace is safe in all cases. 

As illustrated in this [article](https://csslayout.news/whats-the-difference-between-the-alignment-values-of-start-flex-start-and-self-start/), `(start|end)` and `(flex-start|flex-end)` behave differently in `reverse` contexts.

> `start` and `end` are flow relative, and relate to the writing mode and script direction. The `self-start` and `self-end` values are also flow-relative but they relate to the writing mode and direction of the alignment subject, in this case a flex item.

packages/scss/src/commons/core.scss
```diff
@@ -13,8 +13,10 @@ $gaps: 'column-gap', 'row-gap', 'gap';
 $displays: 'block', 'flex', 'inline', 'inline-flex', 'inline-block', 'contents', 'none';
 $direction: 'column', 'column-reverse', 'row', 'row-reverse';
 $wrap: 'wrap', 'nowrap', 'wrap-reverse';
-$justify: 'flex-start', 'flex-end', 'space-around', 'space-between', 'space-evenly', 'center';
-$align: 'flex-start', 'flex-end', 'baseline', 'stretch', 'center';
+// flex-start and flex-end are deprecated
```
fbasmaison-lucca, line 16: They are not really deprecated, just specific to flex boxes. `start` will default to `flex-start` in the context of a flex container, but `flex-start` will not always give the same result as `start` in every context.

image

image

vvalentin-lucca:
@fbasmaison-lucca Thanks for the insight. I'm re-drafting this PR before I fully understand it.

PR #3002, opened 2024-08-06
fix(icons): config path

shajz:
## Description

In the `icons` package, there is an import from the `scss` package, which is an error and is also unused. The `icons` package must be independent and this bug prevents LF/icons to be used alone.

The import being unnecessary as no `config` sass variables are used in this file, we simply remove it.

-----

I noticed it while upgrading LF on a project which uses only the `icons` package.

### DIFF
--- a/packages/icons/src/commons/core.scss
+++ b/packages/icons/src/commons/core.scss
@@ -1,7 +1,7 @@
 @use 'sass:string';
 @use 'sass:list';
 
-@use '@lucca-front/scss/src/commons/config';
+@use '@lucca-front/icons/src/commons/config';
 
 @function capitalize($string) {
 	@return string.to-upper-case(string.slice($string, 1, 1)) + string.slice($string, 2);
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3002/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21406/ui-diff)

jeremie-lucca:
It seems to be a typo while we replaced the import by the full path 2 years ago, but it looks like this import wasn't use, so I guess we can get rid of it.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3002/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21428/ui-diff)

jeremie-lucca (approved):

PR #3003, opened 2024-08-06
chore(lint): let prettier config handle template formatting alone

Supamiu:
## Description

Having both prettier and @angular-eslint configs for template formatting was causing conflicts and issues in the result template. Removing the angular-eslint part from settings fixed that, as well as updating lint-related deps.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3003/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21412/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3003/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21413/ui-diff)

.eslintrc
```diff
@@ -1,160 +1,153 @@
...
-        "@angular-eslint/no-host-metadata-property": "off"
-      }
-    },
-    {
-      "files": [
-        "*.html"
-      ],
-      "extends": [
-        "plugin:@angular-eslint/template/recommended",
-        "plugin:prettier/recommended"
```
shajz, line 139: We use this exact configuration in other projects and it doesn't raise any issue. I don't understand why removing `prettier/recommended` here solves anything 🤔

Supamiu: Me neither, I tested on several files and it makes no sense.

If I add it back, the result format is derping, if I revert all the changes and exclude eslint for formatting (and thus format with only prettier), it works fine. Now if I enable prettier only in this eslint config, I get an error due to `<` being unexpected, showing that somehow it's considering that as a ts file???

shajz: Isn't the problem that 
- in prettier config : `.component.html` and `.page.html` files are used with the angular parser
- in prettier config : `.html` files are used with the html parser
- in eslint config : all `.html` files are used with the same rules ?

Maybe there should be the same split between files in eslint config ? 🤔 and `.html` files would just use `prettier:prettier/recommended` while `.component.html` and `.page.html` files would use both angular-eslint and prettier ?

Or maybe just use `.html` in prettier like in the recommended config by angular eslint : https://github.com/angular-eslint/angular-eslint/blob/main/docs/CONFIGURING_ESLINTRC.md#notes-for-eslint-plugin-prettier-users

PR #3022, opened 2024-08-14
feat(forms): new phone number input component

Supamiu:
## Description

This PR adds a new input for phone numbers which makes use of [libphonenumber-js](https://www.npmjs.com/package/libphonenumber-js) to validate, parse and format phone numbers.

It exposes:

- The input itself.
- A `PhoneNumberFormatter` class that has a `format` static method (should we just export the function and ditch the class?), used as proxy to `libphonenumber-js`'s `format` implementation, handling many formats including national and international.
- A `luPhoneNumberFormat` pipe exposing the formatter as a pipe for template formatting.
- A `PhoneNumberValidators` with a `validPhoneNumber` static method (this one is in a class to keep the `Validators.something` syntax as much as possible) to be used as validator.
- `E164Number`, `CountryCode`, `NumberFormat` and `ValidatePhoneNumberLengthResult` types from `libphonenumber-js` so you don't have to add a dependency to it to properly type your code using phone number input.

🚧 **TODO**:

- [x] We need to plug in aria attributes to make sure this is ok for screen readers (@vvalentin-lucca)
- [x] HTML/CSS integration needs to be properly done, atm it's two inputs and that's it
- [ ] Should we revert format to raw value on focus for UX?

-----

Kudos to @t8g for the nice pair session allowing us to challenge the design properly 🙏 

-----


jeremie-lucca:
Great job!
I identified 46 missing flags. I will start to create them and if I miss time before the release I'll add temporary placeholder including the country ISO Alpha 2: 
image
By checking other Phone number components, it looks they tend to use regional flags on cases like Martinique (which is a good thing imho, plus, it helps to differentiate options entries). 

I'll take the basic UI integration and let the accessibility part to @vvalentin-lucca.

In my opinion, it's preferable to revert raw value on focus (like Number Format).

@Supamiu Would it possible to improve the search on country code? (E.g. being able to select France by searching `+33`, `33` or `FR`) 

Also, we have light performance issues which are worsen a bit by adding flags on the dropdown. Do we have a way to improve it (by using pagination for example).

Supamiu:
We definitely need pagination or virtual scrolling for this and yeah, search should be added to the TODOs too 😸

Supamiu:
I added search and flags, virtual scrolling is going to be a big thing tho as it'd have to be done on simple-select's panel itself, so we'll have to design this properly and make sure we want it, ping @GuillaumeNury

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3022/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21831/ui-diff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3022/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21835/ui-diff)

packages/ng/forms/public-api.ts
```diff
@@ -10,6 +10,7 @@ export * from './switch-input/switch-input.component';
 export * from './radio-group-input/radio-group-input.component';
 export * from './radio-group-input/radio/radio.component';
 export * from './textarea-input/textarea-input.component';
+export * from './phone-number-input/index';
```
GuillaumeNury, line 13: We should move this component in its own lib to avoid adding libphonejs deps to all codebases. We could import it from `@lucca-front/ng/forms/phone-number-input`.

packages/ng/forms/phone-number-input/formatter.ts
```diff
@@ -0,0 +1,8 @@
+import { E164Number, NumberFormat } from './types';
+import { parsePhoneNumber } from 'libphonenumber-js';
+
+export class PhoneNumberFormatter {
```
GuillaumeNury, line 4: I would expose the function directly. Maybe we could call it `luParsePhoneNumber` to avoid conflict with auto-import

packages/ng/forms/phone-number-input/phone-number-input.component.ts
```diff
@@ -0,0 +1,146 @@
...
+	touched(): void {
+		this.#onTouched?.();
+	}
+
+	formatNationalNumber(): void {
+		// TODO: Should we revert format on focus for better edit UX?
+		try {
+			if (isValidPhoneNumber(this.nationalNumber, this.prefixEntry.country)) {
+				this.nationalNumber = this.parsedPhoneNumber?.formatNational();
+			}
```
GuillaumeNury, line 137: Can this code throws ?

Supamiu: It can't because we're making sure the phone number is valid before calling it.

packages/ng/forms/public-api.ts
Supamiu: I'm not a big fan of having such deep imports but you're right, it shouldn't be in all bundles if they're not using the phone number input

packages/ng/forms/phone-number-input/formatter.ts
Supamiu: What if we expose just a function that takes the number and the format? basically removing the class that hosts it.

I'm not a big fan of exposing the parse method with our own name because it'd just mean rebranding it basically while we could expose it as-is just like we do for types. But also, exposing a simple way to format without having to call the parsing by yourself seems nice to have, idk

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3022/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/21854/ui-diff)

GuillaumeNury (approved):

jeremie-lucca (approved):

PR #3053, opened 2024-09-03
Fix ellipsis calculation

vvalentin-lucca:
## Description

Calculation with a clone posed problems because the clone was injected into a `div`, itself injected into the current context, which posed structural problems.

Moving the clone into the body, on the other hand, posed inheritance problems.

So we've changed our approach: we now duplicate the properties rather than the element.

-----

-----


Supamiu (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/22338/ui-diff)

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/22381/ui-diff)

Supamiu (approved):

fbasmaison-lucca (approved):

PR #3058, opened 2024-09-05
Fix form field margins

vvalentin-lucca:
## Description

-----

-----

Before:
Capture d’écran 2024-09-05 à 10 36 17

After: 
Capture d’écran 2024-09-05 à 10 45 18

### DIFF
--- a/packages/scss/src/components/form/component.scss
+++ b/packages/scss/src/components/form/component.scss
@@ -8,6 +8,10 @@
 			margin: 0;
 			padding: 0;
 			border: 0;
+
+			.form-field {
+				margin-block: var(--pr-t-spacings-50);
+			}
 		}
 	}
 }


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3058/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/22434/ui-diff)

packages/scss/src/components/form/component.scss
```diff
@@ -8,6 +8,10 @@
 			margin: 0;
 			padding: 0;
 			border: 0;
+
+			.form-field {
+				margin-block: var(--pr-t-spacings-50);
```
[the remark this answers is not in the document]
vvalentin-lucca: In this case, the container is always a `.form-fieldset`, and its display shouldn't move.

jeremie-lucca, line 13: Applying only a margin bottom on `.formLabel` and `.form-field` wouldn't be cleaner?

vvalentin-lucca: That would work too, but it affects two elements instead of one.
The margins merge here, is that what's bothering you?

jeremie-lucca: I just wonder if margins collapse works well on every browsers

jeremie-lucca (approved):

PR #3087, opened 2024-09-24
ref: handle namespaced at-root and :root selectors

GuillaumeNury:
## Description

Icons ans scss can now be imported inside any element without polluting the global style.

-----

-----

### DIFF
--- a/packages/scss/src/commons/utils/namespace.scss
+++ b/packages/scss/src/commons/utils/namespace.scss
@@ -2,7 +2,7 @@
 
 $defaultAtRoot: 'without: rule';
 
-@if ($isNamespaced) {
+@if (config.$isNamespaced) {
 	// When namespaced, we need to keep the surrounding selector
 	$defaultAtRoot: 'with: rule';
 }
--- a/packages/scss/src/commons/utils/namespace.scss
+++ b/packages/scss/src/commons/utils/namespace.scss
@@ -1,8 +1,11 @@
 @use '@lucca-front/scss/src/commons/config';
 
-// default to 'without: rule' is namespace.$isNamespaced is false
-// default to 'with: rule' is namespace.$isNamespaced is true
-$defaultAtRoot: config.$isNamespaced and 'with: rule' or 'without: rule';
+$defaultAtRoot: 'without: rule';
+
+@if ($isNamespaced) {
+	// When namespaced, we need to keep the surrounding selector
+	$defaultAtRoot: 'with: rule';
+}
 
 @mixin appendRootVars {
 	@if config.$isNamespaced {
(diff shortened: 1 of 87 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3087/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/23142/ui-diff)

packages/scss/src/components/_sample/component.scss
```diff
@@ -2,8 +2,9 @@
 // It does not contain its selector, which is present in the index.
 // Any sub-components are placed in the `@at-root`.
 // The `@at-root` eliminates this selector by default, and its parameter can be used to restore it.
+@use '@lucca-front/scss/src/commons/utils/namespace';
 
-@mixin component($atRoot: 'without: rule') {
+@mixin component($atRoot: namespace.$defaultAtRoot) {
```
vvalentin-lucca, line 7:
```suggestion
@mixin component($atRoot: config.$atRoot) {
```

I'd like to see something like that here.

packages/scss/src/commons/utils/namespace.scss
```diff
@@ -0,0 +1,17 @@
+@use '@lucca-front/scss/src/commons/config';
+
+// default to 'without: rule' is namespace.$isNamespaced is false
+// default to 'with: rule' is namespace.$isNamespaced is true
```
fbasmaison-lucca, line 4:
```suggestion
// default to 'without: rule' if namespace.$isNamespaced is false
// default to 'with: rule' if namespace.$isNamespaced is true
```

packages/scss/src/commons/utils/namespace.scss
```diff
@@ -0,0 +1,17 @@
...
+$defaultAtRoot: config.$isNamespaced and 'with: rule' or 'without: rule';
```
fbasmaison-lucca, line 5: Having the need to explain what the code does seems like a signal it could be simplified.

The oneliner is interesting to understand Sass specificities, but isn’t it slowing down understanding compared to the following example?

```scss
$defaultAtRoot: 'without: rule';

@if $isNamespaced {
  $defaultAtRoot: 'with: rule';
}
```

packages/scss/src/components/_sample/component.scss
GuillaumeNury: `$defaultAtRoot` is not configurable. Should it be moved in `config` anyway ?

vvalentin-lucca: It's not configurable, but it can be modified, so I found default in the name confusing on first reading.
`$globalAtRoot` perhaps?

In config, you have everything you need to configure LF, and if this configuration is modifiable by the BUs, the addition of `!default` at the end of the declarations.

So, to sum up, I'd say :
 - it goes in the config file
 - it doesn't take `!default`

vvalentin-lucca (dismissed):

vvalentin-lucca (approved):

PR #3125, opened 2024-10-09
[emptyState] maxWidth + sticky

### LINKED ISSUE #3096: [Table / Index Table] Empty state & overflow
When Tables have horizontal scroll, empty state can appears offscreen as it based on `100rem`'s container max width (`--components-container-max-width`).  

We could fix it by overriding this value in `.table .emptyState` & `.indexTable .emptyState` context (maybe with a calculation in `vw` and substracting navside width) 

image

vvalentin-lucca:
## Description

Fixes #3096 

- max container width is now based on viewport width
- the component is set to sticky so that it is always visible if horizontal scrolling is present
- two CSS variables can be used to adjust the maximum size and position to the current context (for example, with a compact navSide)

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3125/storybook

PR #3136, opened 2024-10-14
fix(user-display): handle `null` string in isNotEmptyString

Supamiu:
## Description

Fix `Cannot read properties of undefined (reading 'length')` from datadog report.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3136/storybook

GuillaumeNury (approved):

PR #3143, opened 2024-10-16
feat: new DecorativePalette type for when they can be used as palette

Supamiu:
## Description

Add a new type for decorative palette so the components that should handle them can, such as tag.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3143/storybook

jeremie-lucca (approved):

PR #3154, opened 2024-10-23
[Grid] Revert grid width management

jeremie-lucca:
## Description

We need to revert grid width management as it breaks UI on some grid/flex contexts (like horizontal fieldset).

-----

@fbasmaison-lucca We could reconsider this fix after the release

-----


vvalentin-lucca (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3154/storybook

PR #3212, opened 2024-11-14
[statusBadge] wrapper

vvalentin-lucca:
## Description

-----

-----


packages/ng/statusBadge/statusBadge.component.html
```diff
@@ -0,0 +1 @@
+{{label}}
```
Supamiu, line 1: Can we have something else that just a string as content? Maybe we should use `ng-content` to handle that in a more intuitive fashion?

Also, one-liner templates are better as inline `template` rather than in their own file.

packages/ng/statusBadge/statusBadge.component.ts
```diff
@@ -0,0 +1,30 @@
...
+	},
+})
+export class StatusBadgeComponent implements OnChanges {
+	#luClass = inject(LuClass);
+
+	@Input({ required: true })
+	label: string;
+
+	@Input()
+	size: 'M' | 'L' = 'M';
```
Supamiu, line 22: Setting size to default `M` means that parent components won't be able to change that using just CSS, with the fact that we're checking for the value itself in `ngOnChanges`, I wouldn't set a default size here.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3212/storybook

stories/documentation/texts/status-badge/status-badge-angular.stories.ts
```diff
@@ -0,0 +1,49 @@
...
+function getTemplate(args: StatusBadgeAngularStory): string {
+	const size = args.L ? `size="L"` : '';
+	const pal = args.palette ? `palette="${args.palette}"` : '';
+	return `<lu-status-badge ${pal} ${size} label="${args.label}" />`;
+}
+
+const Template: StoryFn<StatusBadgeAngularStory> = (args) => ({
+	props: args,
+	template: getTemplate(args),
+});
```
Supamiu, line 46: If we want to keep `label` as input for the content, this story could be made better by using the native `component` parameter 🙂

vvalentin-lucca: Looking for a little help on this subject.

jeremie-lucca (approved):

PR #3213, opened 2024-11-15
fix(multi-select): fix errors display and aria stuff on displayer

Supamiu:
## Description

We were using the wrong Input directive, resulting in `luInput` missing from the native input, thus breaking links with error manegement and label aria.

Reported by @dclary-lucca 

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3213/storybook

shajz (approved):

PR #3218, opened 2024-11-19
[Select simple & multi] Critical color missing on label / inline message in error state

jeremie-lucca:
## Description

Critical color missing on label / inline message in error state

-----

### DIFF
--- a/packages/scss/src/components/form/index.scss
+++ b/packages/scss/src/components/form/index.scss
@@ -56,7 +56,9 @@
 	}
 
 	&:has(.textField-input-value[aria-invalid='true']),
-	&:has(.timePicker-fieldset-group-textfield-input[aria-invalid='true']) {
+	&:has(.timePicker-fieldset-group-textfield-input[aria-invalid='true']),
+	&:has(.simpleSelect-input[aria-invalid='true']),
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3218/storybook

packages/scss/src/components/form/index.scss
```diff
@@ -56,7 +56,9 @@
 	}
 
 	&:has(.textField-input-value[aria-invalid='true']),
-	&:has(.timePicker-fieldset-group-textfield-input[aria-invalid='true']) {
+	&:has(.timePicker-fieldset-group-textfield-input[aria-invalid='true']),
+	&:has(.simpleSelect-input[aria-invalid='true']),
+	&:has(.multipleSelect-displayer-search[aria-invalid='true']) {
```
vvalentin-lucca, line 61: `[aria-invalid='true']` without a prefix would be too permissive?

vvalentin-lucca (approved):

packages/scss/src/components/form/index.scss
jeremie-lucca: I wonder... It could be a good approach but I'm a bit afraid of specific cases and it would require extra tests.

PR #3224, opened 2024-11-21
fix(multi-select): emit mode none when select all then unselect each

GuillaumeNury:
## Description

Each time a width is read after a class has been added or the dom has been updated, the browser recompute all layers (reflow). Before this PR, this occured **twice per tooltip**.
Now, this append twice, whatever number of tooltip is appended at the same time.

Before:
image

After:
image

-----

-----


GuillaumeNury:
wrong branch...

PR #3236, opened 2024-11-25
fix(date-input): properly handle min/max in a month picker configuration

Supamiu:
## Description

We used to compare month value, which was bad because months are the same accross years, obviously...

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3236/storybook

GuillaumeNury (approved):

Issue #3233, opened 2024-11-22
bug(form-field): nested form-fields make parent consider all controls as children


Supamiu:
## Description

When using a `lu-form-field` inside another `lu-form-field` with each having its own control (like in a radio setup for instance), because `lu-form-field` gets its children `NgControl` references using a `@ContentChildren(NgControl, { descendants: true })`, it'll get controls that are under its scope but also the one under all the possible `lu-form-field` children.

Because we want to allow controls to be nested in possible wrappers or even just a div, we can't disable `descendant` in the query.

## Steps to reproduce

Here is an example template that reproduces this bug:

```html
<form [formGroup]="form">
	<lu-form-field label="Level 1">
		<lu-radio-group-input formControlName="foo">
			<lu-radio [value]="0">
				Pas Coucou
			</lu-radio>
			<lu-radio [value]="1">
				Coucou
				@if(form.value.foo === 1) {
					<lu-form-field label="Level 2">
                                                <!-- This one is a required control -->
						<lu-text-input formControlName="bar"></lu-text-input>
					</lu-form-field>
				}
			</lu-radio>
		</lu-radio-group-input>
	</lu-form-field>
</form>
```

Bug can also be reproduced in `chore/repro-3233` branch, using the `Documentation/Forms/Fields/Form Field/Nested Controls` story that uses the above template.

Selecting `Coucou` radio will mark `Level 1` label as required while it shouldn't, because its control isn't required, only the inner child control is.

## Possible solutions

What we want is a way for `lu-form-field` to know it shouldn't consider a given control as its own control because it's already handled, the two solutions I can see so far are:

- Add a meta value on control to mark it as "handled" by a given control, which should be fine but is, imo, prone to race conditions.
- When getting a control from the query, check its parents recursively until we find a `lu-form-field`, then compare it with current field, if it's the same, we can consider it as this field's control, else no.

lmullot:
On a toujours le problème chez nous, c'est même aggravé car là sans le `[invalid]="false` on a toujours des erreurs alors qu'on n'est pas censé en avoir.

lmullot:
On a trouvé comment fix en local je reclos

PR #3246, opened 2024-12-02
refactor(form-field): signals everywhere !

Supamiu:
## Description

Refactored `lu-form-field` to use signals in many places, fixing numerous change detection-related issues, including https://github.com/LuccaSA/lucca-front/issues/3233.

-----

This will require a quite big technical review to make sure we didn't break anything unwillingly, all my tests were fine but still.

-----

### DIFF
--- a/packages/ng/form-field/form-field.component.ts
+++ b/packages/ng/form-field/form-field.component.ts
@@ -66,7 +66,7 @@ export class FormFieldComponent implements OnDestroy {
 	 */
 	hiddenLabel = input(false, { transform: booleanAttribute });
 
-	rolePresentationLabel = input(false, { transform: booleanAttribute });
+	rolePresentationLabel = model(false);
 
 	inline = input(false, { transform: booleanAttribute });
 
--- a/packages/ng/core/portal/portal.directive.ts
+++ b/packages/ng/core/portal/portal.directive.ts
@@ -84,10 +84,6 @@ export class PortalDirective<T = unknown> implements OnChanges, OnDestroy {
 		}
 	}
 
-	public static ngTemplateContextGuard<T>(_dir: PortalDirective<T>, ctx: unknown): ctx is void {
-		return true;
-	}
-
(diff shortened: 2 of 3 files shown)


packages/ng/form-field/form-field.component.ts
```diff
@@ -30,130 +31,87 @@ let nextId = 0;
...
 	 * Max amount of characters allowed, defaults to 0, which means hidden, no maximum
 	 */
-	@Input()
-	counter = 0;
-
-	required = false;
+	counter = input<number>(0);
 
 	get contentLength(): number {
 		return (this.#inputs[0]?.host?.nativeElement as HTMLInputElement)?.value.length || 0;
```
GuillaumeNury, line 117: If `#inputs` becomes a `Signal`, we could have a `computed` for `contentLength` too

packages/ng/form-field/form-field.component.ts
```diff
@@ -183,6 +141,20 @@ export class FormFieldComponent implements OnChanges, OnDestroy, DoCheck {
...
+			this.updateAria();
+		});
+
+		effect(() => {
+			this.#luClass.setState({
+				[`mod-${this.size()}`]: !!this.size(),
+				'mod-checkable': this.layout() === 'checkable',
+				'form-field': this.layout() !== 'fieldset',
+			});
+		});
```
GuillaumeNury, line 155: this is satisfaying!

packages/ng/form-field/form-field.component.ts
Supamiu: I don't think we can make it a signal due to how it's built and updated

GuillaumeNury: Indeed, we would have to listen to input/change event

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3246/storybook

Issue #3235, opened 2024-11-25
bug(lu-number-input): `min` & `max` validator initlialization race condition


Supamiu:
When using a `lu-number-input` with `min` and/or `max` input, they are forwarded to the inner `input type="number"` which triggers Angular's [MinValidator](https://github.com/angular/angular/blob/1bdc3f1aaef5a7e8bee25447aa8619872cf0eb96/packages/forms/src/directives/validators.ts#L281) or the corresponding for max.

This produces a changes call (https://github.com/angular/angular/blob/1bdc3f1aaef5a7e8bee25447aa8619872cf0eb96/packages/forms/src/directives/validators.ts#L158) which doesn't trigger a new validation cycle because it happens before `registerOnValidatorChange` which is declared few lines below.

This means that if a value is already present when adding the field, it won't be checked by the newly added validators, resulting in form being marked as valid despite being invalid.

Further investigation is needed but a quick and dirty fix on the consumer end is to add a new `patchValue` call after initializing the form.

lmullot:
Ok found why I had an issue with the min / max validators. Values used were not computed unless I made an update on the input value.
Sorry !

PR #3251, opened 2024-12-04
fix(number-input): fix tentative for min & max validators

Supamiu:
## Description

I can't reproduce the bug properly so it's hard to say if the fix is really working :/

#3235

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3251/storybook

GuillaumeNury:
@Supamiu should we consider refacto to signal input something breaking?

Supamiu:
It's a big topic indeed, we might want to keep this one for 19.1

jeremie-lucca (approved):

PR #3254, opened 2024-12-04
Apply focus style by default on buttons

jeremie-lucca:
## Description

Apply focus style by default on buttons

-----

### DIFF
--- a/packages/scss/src/commons/base.scss
+++ b/packages/scss/src/commons/base.scss
@@ -26,7 +26,8 @@
 		@if config.$fontFamily != 'Source Sans Pro' {
 			@font-face {
 				font-family: '#{config.$fontFamily}';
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3254/storybook

vvalentin-lucca (approved):

packages/scss/src/commons/base.scss
```diff
@@ -147,6 +155,12 @@
 			}
 		}
 
+		button {
```
AlexiaAlarza, line 158: pas possible de mettre la class "menu-list-item-action" ?

jeremie-lucca: That was our first idea, but the focus style is applied by default on every `<a>`, so it seemed more consistent to do the same thing on `<button>` (and we're not taking huge risks by automatically replacing default outline browser by ours.)

PR #3260, opened 2024-12-05
feat(select): add required validator for multiselection control

lukah:
## Description

Implemented the `required` validator for `LuMultiSelection` controls, which checks if the selection mode is set to 'none' and returns a validation error if true.

-----

### DIFF
--- a/packages/ng/multi-select/select.validators.ts
+++ b/packages/ng/multi-select/select.validators.ts
@@ -1,0 +1,13 @@
+import { AbstractControl, ValidationErrors } from '@angular/forms';
+import { LuMultiSelection } from './select.model';
+
+const required = (control: AbstractControl<LuMultiSelection<unknown>>): ValidationErrors | null => {
+	if (!control.value || control.value.mode === 'none') {
+		return { required: true };
+	}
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3260/storybook

packages/ng/multi-select/select.validators.ts
```diff
@@ -0,0 +1,13 @@
+import { AbstractControl, ValidationErrors } from '@angular/forms';
+import { LuMultiSelection } from './select.model';
+
+const required = (control: AbstractControl<LuMultiSelection<unknown>>): ValidationErrors | null => {
+	if (control.value && control.value.mode === 'none') {
```
GuillaumeNury, line 5:
```suggestion
	if (!control.value || control.value.mode === 'none') {
```

should we treat `null` value as well?

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3260/storybook

GuillaumeNury (approved):

PR #3281, opened 2024-12-11
TooltipTriggerDirective: change inputs for models

Ptit-Biscuit:
## Description

Change `input`s for `model`s to set properties from outside.

-----

Th goal is to be able to set `luTooltip`, `luTooltipDisabled` etc directly within a `.ts` file. E.g. on T.Abs we have a `translateI18n` directive that adds a tooltip on host so we want to set it within `ngOnInit`.

-----

### DIFF
--- a/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
+++ b/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
@@ -14,7 +14,7 @@ import {
 	computed,
 	effect,
 	inject,
-	input,
+	model,
 	numberAttribute,
 	signal,
 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


Supamiu:
@GuillaumeNury I was thinking about this yesertday and I thought: what about using `linkedSignal`s for input overrides?

Something like:

```ts
luTooltip = input<PortalContent>();

tooltipContent = linkedSignal(() => this.luTooltip());
```

This would allow overriding the value from any component using this as hostDirective without using a `model`, additionally, with a proper naming convention, we could make it easy to use for most devs.

However, it wouldn't fix the issue with required imports, but that's on Angular.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3281/storybook

PR #3333, opened 2025-01-07
fix(form-field): async loading for controls detection + test story

Supamiu:
## Description

Because we were refreshing controls based on the new `events` observable, it's not emitting anything until an actual event is pushed in the Observable.

I solved this by just adding an initial start value so we can have the control's initial state.

Also added a story for testing purpose.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3333/storybook

thomasmarcelin (approved):

jeremie-lucca (approved):

PR #3349, opened 2025-01-21
[Callout] Required / optional doc

jeremie-lucca:
## Description

Add required / optional info on Description / Heading

-----


jeremie-lucca:
@Supamiu We're trying to avoid usage of headings without description. Would it be possible to add a warning if `lu-callout`'s description is empty?

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/27949/ui-diff)

Supamiu:
We can't really do this without impacting performances, as the content is projected using `ng-content`.

jeremie-lucca:
Ok, let's keep only the SB doc, then

vvalentin-lucca (approved):

PR #3358, opened 2025-01-22
fix/flexBasis

vvalentin-lucca:
## Description

The `flex-basis: 1` was missing.

-----

-----


packages/scss/src/commons/core.scss
```diff
@@ -17,7 +17,7 @@ $wrap: 'wrap', 'nowrap', 'wrap-reverse';
 $justify: 'flex-start', 'flex-end', 'space-around', 'space-between', 'space-evenly', 'center';
 $align: 'flex-start', 'flex-end', 'baseline', 'stretch', 'center';
 $flex: '0', '1';
-$basis: '0', 'auto';
+$basis: '0', '1', 'auto';
```
fbasmaison-lucca, line 20: Wouldn’t a value other than `0` or `auto` need a unit?

PR #3396, opened 2025-02-06
fix(multi-select): wrong selection/unselection when manually selectin…

### LINKED ISSUE #3352: [Multi Select][Establishment] Weird group selection behavior
When selecting / deselecting establishment groups, the wrong group can be selected / deselected.

https://github.com/user-attachments/assets/bfc749ca-f66b-479b-95e3-6707947ff168

GuillaumeNury:
## Description

When manually selecting all options, the select internally store a list with all options instead of relying on selectAll flag. Hence, when unchecking one option, it would keep only this option instead of unchecking it.

Fix #3352

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3396/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/28998/ui-diff)

PR #3409, opened 2025-02-11
fix(form-field): Use invalid input value

3htomit:
## Description

When using `[invalid]="condition"`, the form field was always in error, as `invalidStatus` computed signal was always returning `true` whatever the `invalid` input boolean value passed.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3409/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/29230/ui-diff)

Supamiu (approved):

PR #3360, opened 2025-01-23
wip

vvalentin-lucca:
## Description

Functional description for the changelog.
Keep an empty line under your text, as well as the 5 lines that follow it.

-----

Optionally, technical or more in-depth description for reviewers.
Keep an empty line under your text, as well as the 5 lines that follow it.

-----

PR #3419, opened 2025-02-13
fix(tooltip): Allow tooltips to break words if they are too long

fbasmaison-lucca:
## Description

If a tooltip is made of a word larger than itself, the content of the tooltip won’t be legible.

The scenario is interesting since we use them for truncated text.

This PR allow words to break in a tooltip.

-----

### Before

Long gibberish word is truncated

### After

Long gibberish word is broken into multiple lines

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3419/storybook

jeremie-lucca:
Already included in https://github.com/LuccaSA/lucca-front/pull/3360 with `hyphens: auto`. It will be available on 19.2 ✌️

PR #3465, opened 2025-02-27
fix(form-field): better required and invalid detection

GuillaumeNury:
## Description

When hot-swapping the control of a `formControl` directive, required/invalid statuses were not updated correctly.

-----

-----


shajz (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/30376/ui-diff)

Supamiu (approved):
Interesting approach that thinks outside of the "full signal" box 👍

PR #3472, opened 2025-03-04
fix(callout-disclosure): use icon based on state for iconless mod

Supamiu:
## Description

We were using the icon input for `mod-iconless` class trigger, meaning that if icon was `null`, we had the class active, thus removing the said icon.
-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3472/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/30632/ui-diff)

GuillaumeNury (approved):

PR #3500, opened 2025-03-17
[loading] Remove redundant code:

fbasmaison-lucca:
## Description

The `L` mixin already includes the `block` mixin, but both are called in other mixins, in the same order.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3500/storybook

jeremie-lucca (approved):

PR #3547, opened 2025-04-03
[footer] buttons with wrap

vvalentin-lucca:
## Description

Now that text in buttons can be returned to the line, we need to correct some behavior in the footer.

-----

-----

Before:

After:


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3547/storybook

jeremie-lucca (approved):

PR #3559, opened 2025-04-09
fix(lu-textarea-input): apply autoresize with initial value

g-drouard:
## Description

Textarea input with default value and autoresize doesn't apply autoresize on init.

-----

Add `startWith` on `valueChanges`.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3559/storybook

PR #3574, opened 2025-04-16
fix(date-range-input): input always disabled

Supamiu:
## Description

Because of a signal used as value instead of being read in template and it being non-`null`, it was always considered as true, hence input being disabled.

We might need a linter rule or something to prevent this in the future.

-----

-----

### DIFF
--- a/packages/ng/date2/date-range-input/date-range-input.component.html
+++ b/packages/ng/date2/date-range-input/date-range-input.component.html
@@ -25,7 +25,7 @@
 					(click)="editedField.set(0);openPopover(popoverRef, 'start')"
 					(keydown.arrowDown)="arrowDown(popoverRef, 'start')"
 					(keydown.escape)="popoverRef.close()"
-					[disabled]="disabled"
+					[disabled]="disabled()"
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3574/storybook

packages/ng/date2/date-range-input/date-range-input.component.html
```diff
@@ -25,7 +25,7 @@
 					(click)="editedField.set(0);openPopover(popoverRef, 'start')"
 					(keydown.arrowDown)="arrowDown(popoverRef, 'start')"
 					(keydown.escape)="popoverRef.close()"
-					[disabled]="disabled"
+					[disabled]="disabled()"
```
shajz, line 28: erf that could have been caught by a lint check :/

shajz (approved):

packages/ng/date2/date-range-input/date-range-input.component.html
Supamiu: > We might need a linter rule or something to prevent this in the future.

Yup :/

PR #3586, opened 2025-04-22
fix(date-input): selecting initial value after a change doesn't reflect on model

Supamiu:
## Description

Because of how we used to compare date with previous write value, resetting the input to its original value from write wasn't resulting in a change event being sent to the control. This has been fixed by resetting value from write call when selected date is being applied.

Also, this PR adds an e2e test to make sure it doesn't come back in the future (and I was able to TDD the fix ✨)

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3586/storybook

GuillaumeNury (approved):

jeremie-lucca (approved):

Issue #3587, opened 2025-04-23
[Rich textfield] The input has autofocus


martincadoux:
When loading a view which contains a rich textfield, it automatically receives focus. This can be verified on [Storybook](https://lucca-front.lucca.io/storybook/iframe.html?globals=&id=documentation-forms-fields-richtextinput-angular--basic&viewMode=story).

PR #3596, opened 2025-04-28
fix(rich-text-input): fixed RTE autofocus when value is set programmatically

obraud:
## Description

Fixed RTE autofocus when value is set programmatically (#3587)

-----

* Disabled autofocus when updating rich text input value automatically
* Removed useless field validity handling since it's already handled by `form-field` and `luInput`

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3596/storybook

PR #3605, opened 2025-04-29
fix(rich-text-input): fixed RTE being marked as touched & dirty when initial value is set

obraud:
## Description

Fixed RTE being marked as touched & dirty when initial value is set

-----

* Added custom lexical update tag, marking `writeValue` udpates to not fire changed/touched events

-----


Supamiu (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3605/storybook

PR #3620, opened 2025-05-09
fix(filter-pills): better handling of nested select inputs

Supamiu:
## Description

When used deep inside the component tree, filter pill inputs might not register properly. This fixes the issue for the one input we've add issues with regarding this: select.

@GuillaumeNury Looks like the inject + register pattern is the solution once more.

-----

We might want to do this for all pills but this is more of a hotfix for now and we'll see how it goes in the future, it wouldn't be breaking for anyone anyways so it can wait a bit.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3620/storybook

jeremie-lucca (approved):

GuillaumeNury (approved):

GuillaumeNury:
@Supamiu whould it be breaking if we drop the "contentChild way" of retrieving the filterpill input ?

PR #3622, opened 2025-05-12
feat(tooltip): allow for external host

### LINKED ISSUE #3541: [Tooltip] Allow to detach tooltip from trigger
In some cases, in would be helpful to : 
- Attach tooltip on a specific element (different from trigger)
- Merge multiple tooltips in a single one

Supamiu:
## Description

This only works properly on non-component elements, just like for popover2, we might want to do it using a directive if any issue appears with that. Reason is that when using `#ref` on a component, it gives us a ref to the component, not the host's `elementRef` that we want.

-----

closes #3541

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3622/storybook

GuillaumeNury (approved):

jeremie-lucca (approved):

PR #3628, opened 2025-05-14
[FileUpload] Fix required

vvalentin-lucca:
## Description

As the `fileUpload` has a special handling, we need to manage its required differently from other inputs. We add an option for this in the `formField`.

-----

Bonus: the `luInputStandalone` didn't seem necessary and correct the `for`/`id` binding.

-----

Capture d’écran 2025-05-14 à 11 12 56


AlexiaAlarza:
🚀

AlexiaAlarza (approved):

Supamiu (approved):

PR #3656, opened 2025-06-02
fix(stylelint): filterBarDeprecated warning

shajz:
## Description

Fixes a stylelint deprecation warning rule regarding `.filterBarDeprecated` and `.filterBarDeprecated-xxx` uses.

The regex is currently reversed so we get a warning for all `.filterBar` and `.filterBar-xxx` use :(

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3656/storybook

PR #3691, opened 2025-06-18
feat(number-format): currency style -> display a trailing 0 on fraction part if only 1 non-zero digit set

Loaak:
## Description

Add the possibility to keep 2 digits after the comma even if these digits are 0s.

This request is for Angular lu-number-format-input with formatStyle="currency"

-----

### DIFF
--- a/.storybook/public/mockServiceWorker.js
+++ b/.storybook/public/mockServiceWorker.js
@@ -8,7 +8,7 @@
  * - Please do NOT serve this file on production.
  */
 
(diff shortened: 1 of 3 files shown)


mathieu-bellange (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3691/storybook

GuillaumeNury (dismissed):

PR #3699, opened 2025-06-20
fix(date-input): invalid format to date ISO

lmullot:
## Description

Fix an issue on date-input with date-ISO format where non valid string value like 'bug' where missinterpreted and caused an error to be thrown from date-fns librairy.

-----

Prefer checking if date is in valid format before formatting it.

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/37294/ui-diff)

Loaak (approved):
[suggestion]
Peut être pour une autre fois, mais ce serait cool d'avoir des test unitaires pour, notamment, ce cas.

Issue #3689, opened 2025-06-18
[date-range] ngModelChange triggered two times at init


batnyu:
When using a `lu-date-range-input` component, `ngModelChange` is triggered with `null` every time.
It is also triggered with the initial value if there is one.

I think `ngModelChange` should be triggered only after a change from the user not if the input in `ngModel` changes. 

https://github.com/user-attachments/assets/bb033c8f-b198-4856-b30b-5cf255db35fa

PR #3664, opened 2025-06-05
feat(date2): allow reset instead of clear

shajz:
## Description

⚠️ This PR target master. Wait 20.0.1 release before merge ⚠️

When date inputs are `clearable` but have an initial default value, the `clear` button could be expected to `reset` the field to its initial value instead of setting it to `null`.

This PR adds a new `clearBehavior` input to both `lu-date-input` and `lu-date-range-input`. It defaults to the current mode (`'clear'`) and accepts a `'reset'` value.

When set to `'reset'`, the clear button will reset the fields to their initial value.

-----

Demo:

```ts
defaultDateRange: DateRange = { start: this.today, end: null };

filterForm = new FormGroup({
  date: new FormControl<Date>(this.today, { nonNullable: true }),
  dateRange: new FormControl<DateRange>(this.defaultDateRange, { nonNullable: true }),
});
filters = controlSignalRawValue(this.filterForm);
isDateFilterClearable = computed(() => !isSameDate(this.filters().date, this.today));
isDateRangeFilterClearable = computed(() => !isSameDateRange(this.filters().dateRange, this.defaultDateRange));
```

```html
<form [formGroup]="filterForm">
  <lu-filter-pill label="Période">
    <lu-date-range-input formControlName="date" clearBehavior="reset" [clearable]="isDateFilterClearable()" />
  </lu-filter-pill>
  <lu-filter-pill label="Période">
    <lu-date-range-input formControlName="dateRange" clearBehavior="reset" [clearable]="isDateRangeFilterClearable()" />
  </lu-filter-pill>
</form>
```

https://github.com/user-attachments/assets/444ebd5b-abf2-49d6-a3be-0e531884edeb

https://github.com/user-attachments/assets/84630315-735d-4078-8927-c072b19f69f1

-----

Notes : The difference in behavior in the `clear()` function in both components is kinda weird. Why does one call `onChange` but the other doesn't? 🤔

PR #3704, opened 2025-06-20
fix(date-range): #3689 removing effect to avoid null value emitted in…

batnyu:
… ngModelChange at init, using onChange in every case we should inform the parent instead, removing setupInputEffect in favor of textInputChange to avoid using an effect, unit tests with spectator

## Description

Closes https://github.com/LuccaSA/lucca-front/issues/3689 

At init, ngModelChange is no longer triggered with null and the initial value if there is one.

-----

It conflicts with https://github.com/LuccaSA/lucca-front/pull/3664.
But the bug is still present on the branch `fix/date2.allow.reset`.

-----

### DIFF
--- a/jest-global-setup.js
+++ b/jest-global-setup.js
@@ -1,0 +1,3 @@
+module.exports = async () => {
+	process.env.TZ = 'UTC';
+};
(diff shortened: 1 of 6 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3704/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/37332/ui-diff)

Supamiu (dismissed):

Supamiu (approved):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/40201/ui-diff)

PR #3757, opened 2025-07-09
Mobile push.enh

bmoriniere:
## Description
I suggest to provide output events on link usage each app can consume according to their own needs. Could be especially useful to track use of links.

-----

-----

### DIFF
--- a/packages/ng/mobile-push/mobile-push.component.html
+++ b/packages/ng/mobile-push/mobile-push.component.html
@@ -7,10 +7,15 @@
 		<div class="mobilePush-content">
 			<div class="mobilePush-content-description"><ng-content /></div>
 			<div class="mobilePush-content-downloadApps">
-				<a [attr.href]="intl.urlAppStore" class="mobilePush-content-downloadApps-link" (click)="appStoreLinkClicked.emit()">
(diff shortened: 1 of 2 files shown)


packages/ng/mobile-push/mobile-push.component.html
```diff
@@ -7,10 +7,15 @@
 		<div class="mobilePush-content">
 			<div class="mobilePush-content-description"><ng-content /></div>
 			<div class="mobilePush-content-downloadApps">
-				<a [attr.href]="intl.urlAppStore" class="mobilePush-content-downloadApps-link">
+				<a target="_blank" [attr.href]="intl.urlAppStore" class="mobilePush-content-downloadApps-link" (click)="appStoreLinkClicked.emit()">
```
bmoriniere, line 10: I don't understand why, but the linter wants inline attributes here and line breaks below 🤔

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3757/storybook

packages/ng/mobile-push/mobile-push.component.html
nsarradin: As it's a target blank link, we need to specify in the image's `alt` that the link is opened in a new tab.
Adding a "(new tab)" at the end of the alt should help

Supamiu: If we want to have an alt, better have it translated tho.

PR #3758, opened 2025-07-10
fix(text-input): fixed missing addon export

ArthurVinbeau:
## Description

Importing `TextfieldIconAddon`, `TextfieldTextAddon` and `TextInputAddon` is not allowed from `@lucca-front/ng/forms/text-input/text-input-addon` because of the moduleResolution: 'bundler' .

We need to use import from `@lucca-front/ng/forms` instead.

-----

fixed missing `text-input-addon` export in `public-api.ts`

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech38744/ui-diff)

Issue #3741, opened 2025-07-02
[Divider] Margins are incorrect in vertical mode


martincadoux:
When using the vertical mode of the Divider component, the 4px block margin should be replaced with an inline margin just like the Figma component.

PR #3762, opened 2025-07-11
fix(divider) mod vertical margins

ymalie-lucca:
## Description

fix: #3741

Both `--components-divider-marginBlock` and `--components-divider-marginInline` rely on the new `--components-divider-margin-value` var which can be easily overridden.

> [!NOTE]
> I'm just not sure about the new variable name, if you have a better idea… 💡 

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3762/storybook

jeremie-lucca (approved):

PR #3778, opened 2025-07-21
fix(dialog): fix typing import issue with dialogs

obraud:
## Description

Fix typing import issue with dialogs

-----

Before the fix, `dialog.d.ts` would contain the following
image
Which breaks target app build when importing `@lucca-front/ng/dialog`

Also there was unnecessary duplication of `ComponentType` type because it was imported from both `@angular/cdk/overlay` and `@angular/cdk/portal`

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech39395/ui-diff)

PR #3792, opened 2025-07-28
fix(core-select): homonyms additionnal data was not added when using search

GuillaumeNury:
## Description

Option was completed too soon. When `getOptionsPage` needs to emit multiple times, it would stop after the first emission when `isLastPage === true` (it is the case with our homonyms management).

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/39904/ui-diff)

PR #3806, opened 2025-08-07
fix(autocomplete): DOM type and off for default

shajz:
## Description

There is a DOM type for the `autocomplete` attribute, we may as well use it!

Also fixes `autocomplete="false"`, it's `off`. https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#value

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/40731/ui-diff)

jeremie-lucca (approved):

PR #3813, opened 2025-08-11
[footer] fix footer sticky without display contents

vvalentin-lucca:
## Description

`display: contents` prevents events from being added to `lu-footer`, so we prefer to move the sticky positioning to the parent.

-----

NB: placing the `.footer` class at the same level as `lu-footer` would have caused an accessibility issue, since `role="contentinfo"` [must only be present on the main footer of the page for RGAA](https://accessibilite.numerique.gouv.fr/methode/glossaire/#zone-de-pied-de-page).

-----

### DIFF
--- a/packages/ng/footer/footer.component.scss
+++ b/packages/ng/footer/footer.component.scss
@@ -1,5 +1,10 @@
 @use '@lucca-front/scss/src/components/footer';
 
-lu-footer {
-	display: contents;
+lu-footer:has(.footer.mod-sticky) {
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3813/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/40925/ui-diff)

PR #3823, opened 2025-08-19
fix(filter-pills): default value - isEmpty + clearable

JeSuisCharlie1:
## Description

Put fix #3821 #3822 in v20-lts.

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/41287/ui-diff)

PR #3849, opened 2025-08-26
fix(user-select): no params for me url

shajz:
## Description

While testing the PR https://github.com/LuccaSA/ilucca/pull/21182, which allows for use of new `uniqueOperationIds` on users entities, I spotted a bug with the `me` user request.

We tried to use specific `search` or `scopedsearch` endpoint while fetching the current user. This is unnecessary and leads to bugs. 

- The current user can be fetched on the `/me` endpoint
- We don't want to apply any search parameters to this request, we only want to know who the current is to remove him from search results if `displayMeOption` is true (by default).

I'm also fixing the case where the API throws an error, returning `null` seems better (dunno why the current user couldn't find himself though 🤷 )

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/41783/ui-diff)

Supamiu (dismissed):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/41978/ui-diff)

PR #3863, opened 2025-09-01
fix(rich-text): add optional parameter when tabindex is set

JeanBeaurepaire:
## Description

Remove console error

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/42300/ui-diff)

Supamiu (approved):

PR #3884, opened 2025-09-08
fix(richtext): tags could be displayed as keys

sledoux-lucca:
## Description

When changing programmatically the content of the richtext, to a content with the same number of tags (typically, by switching language), the tags could be displayed as their keys, not their descriptions.
Additionnally, the tags were disabled in this case.

Base state (English) : 
image

After changing language : 

Before fix : 
image

After fix : 
image

-----

Missed impacts in the changes of "partial" tag handling : 
- local "history" of tags `#tagNodeKeys` weren't updated in this case, and were still referencing deleted nodes ids.
During the next language change, `newNodes` were then containing old and current nodes, and only then the refresh was triggered (sizes being different). => Comparing values of the sets (ids of Lexical nodes) is needed here.
- missing refactor when removing "partial" attribute => "disabled" was set to true instead of former "partial" attribute.

-----

### DIFF
--- a/packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
+++ b/packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
@@ -199,8 +199,8 @@ function domConversionFunction(domNode: Node): DOMConversion {
 	};
 }
 
-export function $createTagNode(key = '', description?: string, disabled = false): TagNode {
-	return new TagNode(key, description, disabled);
+export function $createTagNode(key = '', description?: string): TagNode {
+	return new TagNode(key, description);
 }
 
 export function $isTagNode(node: LexicalNode | null | undefined): node is TagNode {
--- a/packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
+++ b/packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
@@ -6,7 +6,7 @@ import { LU_RICH_TEXT_INPUT_TRANSLATIONS } from '../../rich-text-input.translate
 import { $createTagNode, TagNode } from './tag-node';
 import type { Tag } from './tag.model';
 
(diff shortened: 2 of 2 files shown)


packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
```diff
@@ -194,8 +198,8 @@ function domConversionFunction(domNode: Node): DOMConversion {
 	};
 }
 
-export function $createTagNode(key = '', description?: string, partial = false): TagNode {
-	return new TagNode(key, description, partial);
+export function $createTagNode(key = '', description?: string, disabled = false): TagNode {
+	return new TagNode(key, description, disabled);
```
JeanBeaurepaire, line 202: disabled n'est plus utilisé dans ce cas

```suggestion
export function $createTagNode(key = '', description?: string): TagNode {
	return new TagNode(key, description, false);
```

packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
```diff
@@ -6,6 +6,8 @@ import { LU_RICH_TEXT_INPUT_TRANSLATIONS } from '../../rich-text-input.translate
 import { $createTagNode, TagNode } from './tag-node';
 import type { Tag } from './tag.model';
 
+const areSetsEqual = (a: Set<unknown>, b: Set<unknown>): boolean => a.size === b.size && [...a].every((value) => b.has(value));
```
JeanBeaurepaire, line 9:
```suggestion
const areSetsEqual = (a: Set<string>, b: Set<string>): boolean => a.size === b.size && [...a].every((value) => b.has(value));
```

JeanBeaurepaire (commented):
Merci beaucoup pour ta PR sur le richtext 👏 🙏

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3884/storybook

JeanBeaurepaire (approved):

Supamiu (approved):

PR #3888, opened 2025-09-11
feat(date): displayMode as model to be able to specify starting mode

Supamiu:
## Description

Add an input called `displayMode` to control starting display mode so we can have it differ from date mode (to start selection with year for instance).

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3888/storybook

PR #3902, opened 2025-09-16
fix(deps): add @angular/animations as peer dependency

Supamiu:
## Description

Even tho it's deprecated, animations are still required for some older features in our lib, we can't assume it's used everywhere anymore.

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/43522/ui-diff)

PR #3930, opened 2025-09-24
feat(rich-text): add support for plain-text plugin

sledoux-lucca:
## Description

For some use-cases with no formatting (typically, mail templates), we'd like to use the Lexical editor in plain-text mode, not rich-text, with other custom LF plugins (Tags in this case).
The most notable difference being the Enter behavior : 
- rich-text : Enter = `\n\n`, Shift+Enter = `\n`
- plain-text : Enter = Shift+Enter = `\n`

The expected result is as below, in the lexical playground :
 
image

image

-----

A new Formatter for PlainText was added, that : 
- registers the lexical plain-text plugin instead of the rich-text plugin (additional npm dependency to be installed by the consumer of this formatter)
- uses the parse/format logic of the markdown formatter, with `shouldPreserveNewLinesInMarkdown` set to true
  - I did this for simplicity, and because the Tags plugin depends on `Transformer` from lexical/markdown anyway
  - to remove the dependency on lexical/markdown for this formatter, i guess we'd have to create the TagNodeTransform logic from scratch, [like in this example](https://lexical.dev/docs/getting-started/creating-plugin#creating-node-transform)

-----


obraud (commented):
Just missing docs for new formater & input parameters  👍

packages/ng/forms/rich-text-input/rich-text-input.component.ts
```diff
@@ -19,6 +19,7 @@ import {
 } from '@angular/core';
 import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 import { createEmptyHistoryState, registerHistory } from '@lexical/history';
+import { registerPlainText } from '@lexical/plain-text';
```
obraud, line 22: It would be ideal to lazyload this, but i'm not sure sure it's actually compatible with all the lifecycle going on in the input component

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/44194/ui-diff)

stories/documentation/forms/fields/rich-text/angular/rich-text-input.stories.ts
JeanBeaurepaire, line 1: Ptet rajouter une story spécifique pour ce use case (markdown ou html)

packages/ng/forms/rich-text-input/formatters/markdown/markdown-formatter.ts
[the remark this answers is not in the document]
sledoux-lucca: Donc normalement y a juste pas de config spécifique supplémentaire coté HTML *

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/44406/ui-diff)

JeanBeaurepaire (approved):

Supamiu (approved):

PR #3945, opened 2025-10-01
fix(tree-select): handle undefined parent

Supamiu:
## Description

Our tree select mapper didn't handle `undefined` in the grouping logic, which is bad due to `Array.find` returning `undefined` when something isn't found.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3945/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/44654/ui-diff)

PR #3947, opened 2025-10-02
feat: Add pattern input to lu-number-input

arnaudoisel:
## Description

Add pattern input to lu-number-input to be able to validate the input against a regular expression.

-----

We need it to resolve this bug in Performance : https://github.com/LuccaSA/Poplee.Talent/issues/10387

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3947/storybook

PR #3956, opened 2025-10-06
fix(departments): simple select optionComparer crash

Supamiu:
## Description

A nullish value would create a crash due to the use of `in` operator on a `null` value.

This has been fixed by returning `null`in the default option key mapper if the option itself is falsy.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-3956/storybook

PR #4011, opened 2025-10-21
[dialog] [select] fix layers

vvalentin-lucca:
## Description

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/46013/ui-diff)

PR #4031, opened 2025-10-27
[layers] add config for each package

vvalentin-lucca:
## Description

The layers were defined in the scss package configuration file. However, this package is not always the first to be called, and not always via its configuration file.

By defining the layers at the beginning of each css file in the three packages, we ensure that regardless of the order in which they are called, the layers will be defined correctly.

-----

-----

### DIFF
--- a/packages/ng/styles/definitions/option/_option-item.scss
+++ b/packages/ng/styles/definitions/option/_option-item.scss
@@ -2,21 +2,22 @@
 @use '@lucca-front/icons/src/commons/utils/icon';
 @use '@lucca-front/scss/src/commons/utils/a11y';
 
+@include namespace.appendRootVars {
+	@layer base {
+		--components-options-item-padding-vertical: var(--pr-t-spacings-50);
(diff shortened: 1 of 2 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/46469/ui-diff)

packages/ng/styles/definitions/option/_option-item.scss
```diff
@@ -2,22 +2,21 @@
 @use '@lucca-front/icons/src/commons/utils/icon';
 @use '@lucca-front/scss/src/commons/utils/a11y';
 
-@include namespace.appendRootVars {
```
jeremie-lucca: It seems to have side effect as some variables are not defined anymore. e.g. in old Establishement: 
image

Group name used to have padding:
image

jeremie-lucca (changes requested):

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/47063/ui-diff)

PR #4057, opened 2025-10-31
feat(core-select): adding occupation-category directive

thomas-lg:
## Description

Adds a new `LuCoreSelectOccupationCategoriesDirective` to provide occupation category selection for simple and multi selects.
It fetches data from the `/organization/structure/api/occupation-categories` endpoint and supports filtering and sorting.

-----

This introduces a new directive for selecting occupation categories. It also adds mock data for the API endpoint in the storybook environment.


Supamiu:
Is there a common need accross Lucca at the moment for this Directive or is it added to LF because it'll be a future need?

alex-delaunay-lucca:
Lucca.Analytics is also interested in this PR. Because we add it to manage the same way as the occupationCategories directive in our filters. [See here](https://github.com/LuccaSA/Lucca.Analytics/pull/844/files#diff-91a1994c3afe7fe1139cad8f890866829eda5262917356bbf0612d88701362ab)

Issue #2992, opened 2024-07-29
[Select] Add maxlength on search inputs


g-drouard:
##Problem

Some APIs are not very happy to see very very long search texts.
I think that functionally there is little point in offering a search field of several hundred or even thousands of characters.

##Solution

Add maxlength of X characters (100 ?) to all search inputs.
I don't think putting the value as an option is very interesting.

PR #4067, opened 2025-11-03
[select] Add max length on api searcher

JeSuisCharlie1:
## Description

issue:https://github.com/LuccaSA/lucca-front/issues/2992

Do we create an input with a default value of 100 to let the user change it ?

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4067/storybook

PR #4093, opened 2025-11-07
[Plg push] Add close option

jeremie-lucca:
## Description

- Add close option
- Click on the close button will hide the component & trigger an event if product teams to store data on cache or database

-----

image

-----


packages/ng/plg-push/plg-push.component.ts
```diff
@@ -1,15 +1,37 @@
...
 	@Input() public heading = '';
+
+	@Input({ transform: booleanAttribute })
+	removable = false;
+
+	@Input({ transform: booleanAttribute })
+	/**
+	 * Is the callout removed? Works with two way binding too.
+	 */
+	removed = false;
```
Supamiu: This should be a `model` imo, because we want to get it as input but also changed (same for removedChange, might want to _remove_ it) and having it as `booleanAttribute` makes no sense considering the model approach.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4093/storybook

CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/47629/ui-diff)

PR #4103, opened 2025-11-12
[tooltip] remove hypens

vvalentin-lucca:
## Description

With hyphenation enabled, they are certainly present when long words need to be truncated, but also in short texts such as here:

Capture d’écran 2025-11-12 à 09 53 56

Capture d’écran 2025-11-12 à 09 45 26

In this PR, I suggest doing this instead: 

Capture d’écran 2025-11-12 à 09 57 12

Capture d’écran 2025-11-12 à 09 57 42

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://qa.lucca.tech/47615/ui-diff)

PR #4131, opened 2025-11-18
feat(simple-select): add prefix support

### LINKED ISSUE #3801: [Select] Customize placeholder
### Intention
In some cases it would be useful to customize placeholder the same way we allow to customize displayer:

https://github.com/user-attachments/assets/58944ff7-1ded-4c10-9ef1-ef617393647c

### Mistakes to avoid

Supamiu:
## Description

This will require some CSS before merging, due to the prefix showing at the end, here is a full screenshot with story selected so you can open it on your end too.

image

-----

closes #3801 

-----


packages/ng/simple-select/input/select-input.component.html
```diff
@@ -1,7 +1,9 @@
 <div class="simpleSelect-field">
-	@if (prefix()) {
-		<span class="simpleSelect-field-prefix"><ng-container *luPortal="prefix()" /></span>
-	}
+	<span class="simpleSelect-field-prefix">
```
Supamiu: This means that even without a prefix, we'd have this span, wouldn't cause issues?

jeremie-lucca (approved):

PR #4176, opened 2025-11-26
fix(dialog): apply canMatch on main route to correctly go to the next sibling in app space

batnyu:
## Description

Now the canMatch works when applied on a dialog route. It will not block the navigation.

-----

We move the canMatch on the main route to correctly search for other sibling routes in app space

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4176/storybook

Supamiu (approved):

PR #4191, opened 2025-11-28
[numberPipe] replace hidden by opacity

vvalentin-lucca:
## Description

The `pr-u-hidden` class did not exist, so it has been replaced by `pr-u-opacity0`, which:
- visually hides the decimals;
- does not remove their space in the layout (in order to preserve alignment);
- does not remove the semantic value of the text (which was a problem with `visibility: hidden`).

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4191/storybook

celeste-gcht (approved):

PR #4261, opened 2025-12-10
[luLink] improve documentation explain navigation handler

JeSuisCharlie1:
## Description

Since we have an external input, the BUs think that we need to use an href with an external URL along with Angular router-link navigation, even though that’s not the case. We need to specify which type of navigation we’re referring to

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4261/storybook

PR #4271, opened 2025-12-11
[luLink] fix url tree generation with link command array

JeSuisCharlie1:
## Description

We were putting an array inside another array, which created an incorrect route behind the scenes.
So if we have a string we put it into an array to create the urlTree otherwise we leave the default value.

-----

-----

### DIFF
--- a/packages/ng/link/link.component.ts
+++ b/packages/ng/link/link.component.ts
@@ -68,8 +68,11 @@ export class LinkComponent {
 
 	@HostListener('click')
 	redirect(): void {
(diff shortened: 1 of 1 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/49828/uidiff)

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4271/storybook

jeremie-lucca (approved):

PR #4285, opened 2025-12-16
[Button] IA + Invert combo

jeremie-lucca:
## Description

Handle IA + invert combo

-----

https://github.com/user-attachments/assets/fac3cb2a-5a3a-4f93-882e-5b4581aae48c

-----

### DIFF
--- a/packages/scss/src/components/button/mods.scss
+++ b/packages/scss/src/components/button/mods.scss
@@ -260,10 +260,8 @@
 }
 
 @mixin invertIA {
-	@include AI;
-	--components-button-AI-background-opacity: 25%;
-	--components-button-backgroundColor: transparent !important;
-	--components-button-color: var(--palettes-neutral-0) !important;
+	--components-button-backgroundColor: transparent;
+	--components-button-color: var(--palettes-neutral-0);
 
 	background-image: linear-gradient(
 		-135deg,
--- a/packages/scss/src/components/button/index.scss
+++ b/packages/scss/src/components/button/index.scss
@@ -8,6 +8,7 @@
 
 	@layer mods {
 		&.mod-AI {
+			&:not(.invert) {
 			@include AI;
 
 			&.mod-invert {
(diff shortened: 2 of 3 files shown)


packages/scss/src/components/button/mods.scss
```diff
@@ -253,8 +253,32 @@
 	&:not(:disabled) {
 		@include color.borderGradient(135deg, var(--palettes-brand-400), var(--palettes-AI-500));
 
-		.lucca-icon {
+		&:not(.mod-invert) .lucca-icon {
```
vvalentin-lucca: We often avoid accumulating `:not`, even if it means redefining .lucca-icon a second time in invertIA. Would that work here?

packages/scss/src/components/button/mods.scss
```diff
@@ -253,8 +253,32 @@
...
+	--components-button-backgroundColor: transparent !important;
+	background-image: linear-gradient(
+		-135deg,
+		color-mix(in srgb, var(--palettes-AI-500) var(--components-button-AI-background-opacity), transparent),
+		color-mix(in srgb, var(--palettes-brand-400) var(--components-button-AI-background-opacity), transparent)
+	);
+
+	--components-button-color: var(--palettes-neutral-0);
+
+	&:hover,
```
vvalentin-lucca: I would create invertIAhover and invertIAactive mixins.

packages/scss/src/components/button/mods.scss
jeremie-lucca: That was my first intention, but it seems a bit overkill as the `button.IA` mixin includes quite a lot.
Maybe I can move the :not selector on index and create 3 separate mixins (AI / invert / invertAI ) ?

CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/50120/uidiff)

JeSuisCharlie1 (approved):

PR #4297, opened 2025-12-19
[filterBar] wrap

vvalentin-lucca:
## Description

There are three different display modes:

- wide: the export button is always in its column on the right (this is the default display), and if there are too many filters, they can wrap to the next line while remaining in their column
- narrow: the export button can move below the filters if there is not enough space; the filters behave as in wide mode
- touch: nothing wraps to the next line, and scrollbox is enabled to display a scrollbar and shadows

-----

-----

wide:
Capture d’écran 2025-12-19 à 10 49 02

narrow:
Capture d’écran 2025-12-19 à 10 49 11

touch:
Capture d’écran 2025-12-19 à 10 49 48


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/50471/uidiff)

PR #4299, opened 2025-12-19
[Tooltip] fix wrong generated id

JeSuisCharlie1:
## Description

Since the host id takes a bit longer to update, we pass the id directly as a component input to give the correct one.
This is non-breaking for BU and only used  this way on our filter bars on the LF side.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4299/storybook

PR #4314, opened 2025-12-23
[Schematics] : Fix running test by mocking ora

JeSuisCharlie1:
## Description

We directly mock in our Jest config instead of doing it in each test like on StackOverflow

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4314/storybook

PR #4315, opened 2025-12-24
[schematics] : Add component path schematics updater

JeSuisCharlie1:
## Description

We create a schematics utils folder for mapping the props to be migrated that are shared between both.
We create a component-path schematic: `ng g @lucca-front/ng:component-path`
To update paths to kebab-case.
I’ve prepared part of the component mapper to later handle the migration of component selectors and names.
can be usefull for prisme package migration.

-----

-----

### DIFF
--- a/packages/ng/scrollBox/ng-package.json
+++ b/packages/ng/scrollBox/ng-package.json
@@ -1,6 +1,0 @@
-{
-	"$schema": "../../../node_modules/ng-packagr/ng-package.schema.json",
-	"lib": {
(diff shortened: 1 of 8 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4315/storybook

CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/50712/uidiff)

PR #4312, opened 2025-12-23
[Schematics] : Add alignment-utilities && update color-text

JeSuisCharlie1:
## Description

Update palettes : `ng g @lucca-front/ng:palettes`
Update alignment utilities : `ng g @lucca-front/ng:alignment-utilities`

-----

-----

PR #4335, opened 2026-01-07
fix(schematics): palettes uses the wrong dash char

shajz:
## Description

In the v21.0.1 release, there is an issue with a wrong dash character in the palettes generator schematic : https://github.com/LuccaSA/lucca-front/pull/4312

-----


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/51165/uidiff)

PR #4341, opened 2026-01-09
fix(link): when external, support navigationExtras and UrlTree

batnyu:
## Description

Adds support for `navigationExtras` and `UrlTree` when using `external` from `luLink`

-----

We could potentially add `relativeTo` by injecting `ActivatedRoute`. But it will ignore the base href set in app.
So we will need something like `location.prepareExternalUrl` but if we specify an absolute path, we will have two base href. So we need a way to know if we need to use this method. Maybe by looking for a slash as first character?

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4341/storybook

JeSuisCharlie1 (approved):

PR #4357, opened 2026-01-13
[dialog] fix double gap without action

vvalentin-lucca:
## Description

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/51552/uidiff)

PR #4359, opened 2026-01-13
[data-table] fix basic stories sticky header hidden arg

JeSuisCharlie1:
## Description

We remove the sticky header argument, we don’t need it in the basic story since we already have a dedicated example in the Sticky / Overflow tab.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4359/storybook

jeremie-lucca (approved):

Issue #4356, opened 2026-01-13
[Highlight Data] Review subtext alignment


AlexiaAlarza:
### If we add subText without bubbles or illustrations, the subText is aligned à right instead of being above value.

<img width="492" height="156" alt="Image" src="https://github.com/user-attachments/assets/26d0eb35-8614-4612-a224-e7274abc176e" />

### How to reproduce
Add subText and put bubbles+illustrations on null

PR #4364, opened 2026-01-15
[highlightData] fix subtext position when there is no illustration

vvalentin-lucca:
## Description

#4356 

-----

-----

Capture d’écran 2026-01-15 à 16 54 21


jeremie-lucca:
We still have an issue if `valueFirst` is `true`:
image

PR #4382, opened 2026-01-20
[callout-disclosure] fix stories no icon by default & select palette options

JeSuisCharlie1:
## Description

Icon is null by default.
Update options for palette select args.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4382/storybook

PR #4397, opened 2026-01-21
chore: fix NodeJS security issue

GuillaumeNury:
## Description

This PR updates the Node.js version go back on track for projects using unsuppored NodeJS versions.

LTS versions of NodeJS are the last three even versions (20, 22 and 24 as of today).
The "current" version is supported as well (ie 25 as of today).

As we use NodeJS only on local and CI side, no need to deploy this pull request once merged.

If this PR builds, it should be good to merge.

-----

## Checklist

- [x] Bump NodeJS to latest LTS version
- [ ] Check that project build

### DIFF
--- a/packages/ng/schematics/lib/local-deps/package-lock.json
+++ b/packages/ng/schematics/lib/local-deps/package-lock.json
@@ -72,6 +72,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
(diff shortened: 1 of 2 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/52159/uidiff)

PR #4435, opened 2026-02-03
[form-field] fix counter content inputs change detection

JeSuisCharlie1:
## Description

Since we switched to OnPush change detection, the component’s input bindings only update once. As a result we go through DoCheck so that it updates on every change.

-----

-----


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/52862/uidiff)

PR #4448, opened 2026-02-05
[formField] fix gap only for checkbox

vvalentin-lucca:
## Description

-----

-----

Before: the gap is reset to 0 when it shouldn't be on the time picker

Capture d’écran 2026-02-05 à 11 51 48

After: the gap is only reset to 0 if the field contains a checkbox

Capture d’écran 2026-02-05 à 11 52 00

### DIFF
--- a/packages/scss/src/components/form/index.scss
+++ b/packages/scss/src/components/form/index.scss
@@ -102,9 +102,11 @@
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


jeremie-lucca (approved):

vvalentin-lucca:
See #4475

PR #4451, opened 2026-02-05
[Button] More flexibily with ng selector

jeremie-lucca:
## Description

Allow to use `luButton` on every HTML elements (as we have `span` cases).

-----


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/53237/uidiff)

PR #4466, opened 2026-02-09
fix: tooltip ellipsis not updating on content change

### LINKED ISSUE #3886: [tooltip] Content change at runtime doesn't trigger `#hasEllipsis` update
## Description

When using `luTooltip` with `luTooltipWhenEllipsis` flag, if the directive is initialized with content that's not ellipsis, it won't update when the content changes and creates an ellipsis.

## Technical explanation

When an open event is received in tooltip trigger, we check the ellipsis status in order to know if we need to open or not (when `luTooltipWhenEllipsis` is enabled).

This checks for the content to  find if we have an ellipsis using the ruler service for that. 

The issue is that when the content changes, nothing is triggered to update this status. We could think that adding this check on mouseenter would be a good idea but the issue would persist for keyboard navigation (because we're changing tabindex value based on if there's a tooltip to open or not).

## Solution

The idea would probably be to connect an observer to the content to update the ellipsis thatus as the content changes even if it's not an input.

Supamiu:
## Description

It's not the perfect solution because we rely on Angular hook which isn't designed for this but the other, probably better approach, would be to use CDK's Observers module, which we cannot do without being breaking because a directive cannot bring in modules as imports and TooltipModule is deprecated in our public API.

-----

closes #3886

-----

### DIFF
--- a/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
+++ b/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
@@ -10,28 +10,29 @@ import {
 } from '@angular/cdk/overlay';
 import { ComponentPortal } from '@angular/cdk/portal';
 import {
-	DestroyRef,
-	Directive,
-	EffectRef,
-	ElementRef,
-	Injector,
-	NgZone,
-	OnDestroy,
-	Renderer2,
+	AfterContentChecked,
 	booleanAttribute,
 	computed,
+	DestroyRef,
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4466/storybook

CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/53177/uidiff)

PR #4476, opened 2026-02-10
[dropdown] fix open position issue

JeSuisCharlie1:
## Description

We couldn’t set a value other than below for the dropdown on the popoverPosition side, so we added an input that lets us know whether we’re in a dropdown context. By default we don’t set anything on this input we check if it’s nil so we can override it.

-----

-----

### DIFF
--- a/packages/ng/popover2/popover.directive.ts
+++ b/packages/ng/popover2/popover.directive.ts
@@ -15,7 +15,6 @@ import {
 	linkedSignal,
 	model,
 	OnDestroy,
-	OnInit,
 	output,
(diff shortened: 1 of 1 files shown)


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/53258/uidiff)

CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/53261/uidiff)

jeremie-lucca (approved):

PR #4529, opened 2026-02-18
fix(lu-box): add (killed) output for [killable]

shajz:
## Description

`lu-box` has a nice `killable` input that allows the user to "kill" the box. But it's not plugged into anything. 

Until now.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4529/storybook

PR #4551, opened 2026-02-27
Fix: RTE being touched on init

sledoux-lucca:
## Description

Context : [slack thread](https://lucca.slack.com/archives/C3W78FWUU/p1772202329863549)

On init, the control attached to a RTE was being automatically touched, creating bugs : 
- error state displayed on init
- "confirm discard changes" dialog poping up even when not interacting with the form

-----

Added a flag `#isRootElementBeingSet` to identify when the callback of `registerUpdateListener` was being called due to `editor.setRootElement()`.

This works reliably because Lexical's update listener fires synchronously inside `setRootElement()` (it calls `commitPendingUpdates` internally), so `#isRootElementBeingSet` is guaranteed to still be true when `#onEditorUpdate` runs, and is reset to false right after.

I didn't find a more intuitive alternative with any public API from Lexical :[

-----

### DIFF
--- a/packages/ng/forms/rich-text-input/rich-text-input.component.ts
+++ b/packages/ng/forms/rich-text-input/rich-text-input.component.ts
@@ -210,18 +210,16 @@ export class RichTextInputComponent implements OnInit, OnDestroy, ControlValueAc
 
 	#onEditorUpdate({ tags, dirtyElements }: UpdateListenerPayload) {
 		const isComposing = this.#editor?.isComposing();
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4551/storybook

CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/54695/uidiff)

obraud (approved):

Issue #4562, opened 2026-03-03
[Rich text] Link edition fails after modification


jeremie-lucca:
### Bug description
When a link is edited, it can be impossible to edit it again from the edition panel

https://github.com/user-attachments/assets/8c92b492-1234-440f-acca-1f72dd64e8fa

### How to reproduce
https://lucca-front.lucca.io/storybook/?path=/docs/documentation-forms-fields-richtextinput-angular--docs
- Click on link
- Edit or remove it
- Create a new link or edit it again
The field is now empty and the link cannot be edited or deleted again

PR #4571, opened 2026-03-04
fix(rich-text-input): fix context issues when editing links

obraud:
## Description

Fix context issues when editing links (#4562).

This fix impacts the following rich text plugins:
* Link tool & Autolink
* Tag plugin

-----

Link tool works as follow:
* get current selection from editor (i.e. caret position)
* open popup with link details
* create, delete or replace link at selection
* Link tool component bears the link popover template. This allows us to easily create the popover using LF components, while allowing us to call commands from the base tool. This template is instanciated & positionned by the link node in itself when displayed for the first time.

Investigation: 

* We found out the bug was only occurring when there were multiple editors on the same page, and only after being edited once via the link popover. In this case, the current selection was null, which should not have been possible as user has to interact with the link to open the popover, setting the selection on the way.
* After first edition, the link **popover** editor context was indeed shifting from the original edited `rich-text-input` instance to the last `rich-text-input` on the page. So when the user tried to edit the link for a second time via the popover, the tool got the current selection from that last editor, which was `null`, hence the empty popup/non-functionnal link deletion.
* That context shifting was only occurring within the scope of the link popover, so this was mostly transparent for the rest of the editor.

What was causing the issue:
* Popover template reference was passed to `LinkNode` class as a `static` value. It was therefore shared between all `LinkNode` instances, in all rich text editors in the current Javascript context. This is problematic when there is more than one editor in this context.
  * When editor A is created, it sets `LinkNode.templateRef` template ref with its own instance (`trA`), and renders existing `linkA` using `trA`.
  * When editor B is created, `LinkNode.templateRef` becomes `trB`.
  * User edits `linkA`. At that point, that link has already been rendered and therefore is still using `trA` in memory, so context is still editor A. However, link is replaced and rerendered using current `LinkNode.templateRef`, which is now `trB`
  * User edits `linkA` again. This time, the template context has shifted to `trB` and editor B. As there is no selection in editor B, the link tool opens an empty popup.

Solution:
* Replacing all static variables in custom nodes with classic object attributes. Angular context should never be shared between instances of Lexical objects. 
* To avoid memory leaks, tag nodes and link nodes are now updating their attached view instead of recreating it. 
* It affects both Link plugin and Tag plugin (even though no issue was identified for the latest).
* As race conditions have been removed from link & tag tool, `afterRenderEffect` has been reverted to an `effect` to avoid unwanted editor focus upon instanciation

-----

### DIFF
--- a/packages/ng/forms/rich-text-input/rich-text-input.component.ts
+++ b/packages/ng/forms/rich-text-input/rich-text-input.component.ts
@@ -1,11 +1,11 @@
 import { CommonModule } from '@angular/common';
 import {
-	afterRenderEffect,
 	booleanAttribute,
 	ChangeDetectionStrategy,
 	Component,
 	computed,
 	contentChildren,
+	effect,
 	ElementRef,
 	forwardRef,
 	inject,
@@ -100,7 +100,7 @@ export class RichTextInputComponent implements OnInit, OnDestroy, ControlValueAc
 	#isRootElementInitialized = false;
 
 	constructor() {
-		afterRenderEffect(() => {
+		effect(() => {
 			if (this.#formField?.presentation() && this.contentPresentation()) {
 				this.#editor?.setRootElement(this.contentPresentation()?.nativeElement ?? null);
 				this.#editor?.setEditable(false);


CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/55112/uidiff)

PR #4584, opened 2026-03-05
fix: Deprecated selectors

fbasmaison-lucca:
## Description

* Some selectors were missing the period making them class selectors.
* String patterns only match full selector, not part of it, which won’t work for combined selectors. We broaden the scope to lint combinations of selectors.
-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4584/storybook

PR #4586, opened 2026-03-05
[fancyBox] Move mod-S in layer

fbasmaison-lucca:
## Description

We can’t use `@layer product` currently.

Moving the mod within `@layer mods` should allow that, and avoid having to use `!important`.

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4586/storybook

PR #4613, opened 2026-03-13
[Push panel] Apply CSS var

jeremie-lucca:
## Description

[Push panel] Apply CSS var

-----

-----


packages/scss/src/commons/base.scss
```diff
@@ -10,6 +10,7 @@
 	color: var(--pr-t-color-text);
 	font: var(--pr-t-font-body-M);
 	text-rendering: geometricPrecision;
+	width: calc(100% - var(--commons-pushPanel-width));
```
vvalentin-lucca:
```suggestion
	inline-size: calc(100% - var(--commons-pushPanel-width));
```

PR #4745, opened 2026-04-10
[dataTable] optimise sticky cols

### LINKED ISSUE #4712: [dataTable] Reset inset-block-* when first or last column is sticky.
### Intention

Position calculations are performed automatically by the Angular wrapper, but in the case of the first and last columns, this isn’t necessary, as they are always set to 0.

### Key points

This would avoid the need for calculations and also allow tables with headers containing colspans to be managed, in cases where only the first and/or last columns are sticky.

vvalentin-lucca:
## Description

-----

Close #4712 

The JavaScript part was suggested to me by an AI; it works and seems logical, but I’m just pointing it out.

-----

### DIFF
--- a/packages/ng/data-table/data-table-cell-header/data-table-cell-header.component.ts
+++ b/packages/ng/data-table/data-table-cell-header/data-table-cell-header.component.ts
@@ -46,7 +46,8 @@ export class DataTableRowCellHeaderComponent extends BaseDataTableCell implement
 	inlineSize = input<string | null>(null);
 
 	insetInlineStart = computed(() => {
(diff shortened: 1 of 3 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4745/storybook

CCNET-iLucca:
[Tests d'interfaces](https://dd.lucca.tech/qa/run/57621/uidiff)

jeremie-lucca (approved):

Supamiu (approved):

Issue #4711, opened 2026-03-31
[indexTable] Inside scrollbox and with pagination


vvalentin-lucca:
### Intention

When an indexTable nested within a scrollBox has pagination, this should always remain visible.

### Key points

Probably with a flex and sticky position.

PR #4811, opened 2026-04-29
feat(indexTable): sticky pagination

ymalie-lucca:
## Description

Allow pagination to be sticky on the Index Table. #4711 
It's the default behavior like the Data Table does, regardless what is used (or not) around the table to manage the scroll.

### Default
Capture d’écran 2026-04-29 à 11 21 01

### With an horizontal scroll
Capture d’écran 2026-04-29 à 11 24 58

### Test
I tried on the basic story, by adding an attribute `style="width:1200px""` on the `<lu-index-table>` element. And also with the use of the `<lu-scroll-box>` around.

-----

### DIFF
--- a/packages/scss/src/components/indexTable/component.scss
+++ b/packages/scss/src/components/indexTable/component.scss
@@ -257,6 +257,8 @@
 			}
 
 			.pagination {
+				position: sticky;
+				inset-inline-end: 0;
 				align-self: flex-end;
 				padding-block: var(--pr-t-spacings-50) 0;
 				padding-inline: 0;
@@ -265,6 +267,13 @@
 		}
 
 		.indexTableWrapper-pagination {
+			display: flex;
+			justify-content: flex-end;
+
+			lu-pagination {
+				display: contents;
+			}
+
 			&:empty {
 				display: none;
 			}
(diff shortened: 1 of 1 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4811/storybook

packages/scss/src/components/indexTable/component.scss
```diff
@@ -265,6 +267,13 @@
 		}
 
 		.indexTableWrapper-pagination {
```
vvalentin-lucca, line 267: You could simply set your position to 'sticky' here, combined with `margin-left: auto` and `inline-size: fit-content`.

ymalie-lucca: Yes that work well too! I first tried to make something similar to the Data Table, but we can do something simpler. I pushed a refacto that only use 3 properties on `.indexTableWrapper-pagination` and nothing else.

```css
position: sticky;
inset-inline-end: 0;
align-self: flex-end;
```

vvalentin-lucca (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4811/storybook

PR #4813, opened 2026-04-29
fix(callout): margin remains visible after removal

batnyu:
## Description

- When a `margin` is applied on `<lu-callout [removable]="true">`, the margin remains visible after clicking the remove button
- Fix host binding reading `removed()` (parent input) instead of `removedRef()` (internal signal), so `[hidden]` was never applied on user click
- Add `&[hidden] { display: none }` to prevent `display: block` from overriding the native hidden attribute

https://github.com/user-attachments/assets/d42a960d-1ec1-41d0-8a2d-03a2760a2f8a

https://github.com/user-attachments/assets/88fe765c-8a39-47db-b13d-56d5f8c1e8ef

-----

### DIFF
--- a/packages/ng/callout/callout/callout.component.scss
+++ b/packages/ng/callout/callout/callout.component.scss
@@ -3,5 +3,9 @@
 @layer base {
 	lu-callout {
 		display: block;
+
+		&[hidden] {
+			display: none;
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4813/storybook

jeremie-lucca (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4813/storybook

PR #4851, opened 2026-05-07
fix(time-picker-part): Fix current value input 

JeSuisCharlie1:
## Description

`hidevalue()` puts "0" in the DOM input even when `isValueSet()` is false (the field appears empty visually via the overlay span). When focus + select() doesn't reliably work, typing 1 appends to the hidden 0 create 10.
don't put value in the DOM input when the field isn't set yet.

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4851/storybook

jeremie-lucca (approved):

Supamiu (approved):

PR #4914, opened 2026-05-26
fix(select): preserve tree indentation inside option groups

batnyu:
## Description

When using `lu-simple-select` with both `[treeSelect]` and `*luOptionGroup`, the option group selector `.lu-picker-content-option-group .optionItem-value` overrides `--components-options-item-padding-horizontal`, which cancels the indentation computed by `--components-treeBranch-level` in tree branches.

This adds a more specific selector for `lu-tree-branch .optionItem-value` inside option groups to restore the tree indentation.

#### Before:
before 1

#### After:
image

-----

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4914/storybook

jeremie-lucca (approved):

PR #4921, opened 2026-05-27
fix(rich-text-input): fix toolbar not appearing when switching off presentation mode

### LINKED ISSUE #4915: [Rich text] Toggling on presentation mode results in empty toolbar
### Bug description

Toggling on presentation mode results in empty toolbar

### How to reproduce

In angular context (not Storybook), switching presentation mode from "false" to "true" results in an empty toolbar (no error message in console).

Image
Image

obraud:
## Description

Fixed toolbar not appearing when switching off presentation mode

-----

Fixes #4915 
firefox_u7jddSfhwK

Cause :
* Form Field uses structural directive to display field in presentation mode or editable mode.
* RTE uses transclusion to register tools to the lexical editor. 
* Angular does not handle conditionnal positionning of the same `<ng-content />`, as it can only be projected once in the parent lifecycle.
* To circumvent that issue, `<ng-content>` was put inside a reusable `<ng-template>`, inserted in both editable & presentation views. Content is projected only when the template is used.
* After first view init, the template is inserted in the editor and content is projected. After presentation mode is switched 'on', the template is transfered to the structural div. Once it's in there, it stays there, regardless of field mode. Hypothesis: content cannot be reprojected once it's in the structural directive, hence resulting in empty toolbar div in editable mode.
* Projected content is bound to the lifecycle of the parent component, so the toolbar angular components themselves were not destroyed, and were still bound to the editor (that's why it was somewhat still functionnal, e.g. link tool).
* This bug was not reproducible with Storybook's native "presentation" input, as it destroys the whole view to change the input. 

Fix:
* As the same `ng-content` can only be projected once, it must be included unconditionally in the editor (i.e. editor root, always visible).  
* Once content is projected, it is stored in a `DomPortal`.
* That portal is then physically transferred in either presentation or editable views. It is therefore never destroyed within the editor's lifecycle.

-----

### DIFF
--- a/packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
+++ b/packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
@@ -1,8 +1,8 @@
 import { ChangeDetectionStrategy, Component, effect, ElementRef, forwardRef, inject, input, OnDestroy, signal, viewChildren, ViewContainerRef } from '@angular/core';
 import { ChipComponent } from '@lucca-front/ng/chip';
 import { intlInputOptions } from '@lucca-front/ng/core';
-import { $getNodeByKey, $getRoot, $getSelection, type Klass, type LexicalEditor, type LexicalNode, type NodeKey } from 'lexical';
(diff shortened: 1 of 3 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4921/storybook

JeSuisCharlie1 (approved):

PR #4935, opened 2026-06-01
fix(core-select): correct totalCount response type for users directive

lmullot:
## Description

Fix the totalCount$ observable in LuCoreSelectUsersDirective to correctly handle the API response structure.

- **Issue**: The observable was incorrectly typed as { count: number } but the API returns { data: { count: number } }
- **Fix**: Updated type annotation and the map function to correctly access the nested count property

-----


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4935/storybook

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4935/storybook

GuillaumeNury (approved):

PR #4939, opened 2026-06-01
[dialog] fix fancy background bubble color

vvalentin-lucca:
## Description

Fallback to the default color when the call is not made via a class.

-----

-----


LucasPat (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4939/storybook

PR #4945, opened 2026-06-02
fix(tooltip): anchor ref and condition

JeSuisCharlie1:
## Description

First bug: on a tooltip anchor that is inside a condition, `luTooltipAnchor` is `undefined` because the variable is either not in the same scope or simply not loaded yet. In that case we use the `host` element itself as the default value. Is that a good solution? It's basically attaching the tooltip directly to the relevant element...

For the second bug, on the `IndexTableRowComponent` side, it is not an `ElementRef`, so `luTooltipAnchor` ends up with `x = undefined` and `y = undefined` causing the tooltip to be positioned in the top-left corner of the page.

One possible solution would be to accept an `ElementRef` and expose it as readonly in `IndexTableRowComponent`. But in that case, would we need to expose it for all of our components that are expected to support this pattern?

-----

https://github.com/LuccaSA/lucca-front/issues/3951

-----

### DIFF
--- a/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
+++ b/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
@@ -20,7 +20,7 @@ import { LuTooltipPanelComponent } from '../panel';
 import { EllipsisRuler } from './ellipsis.ruler';
 
 export interface LuTooltipAnchorRef {
-	elementRef: ElementRef;
+	getElementRef(): ElementRef;
 }
 
 let nextId = 0;
@@ -370,8 +370,8 @@ export class LuTooltipTriggerDirective implements OnDestroy {
 
 		if (isNil(anchor)) {
 			return this.#host;
-		} else if ('elementRef' in anchor) {
-			return anchor.elementRef;
+		} else if ('getElementRef' in anchor) {
+			return anchor.getElementRef();
 		} else {
 			return anchor;
 		}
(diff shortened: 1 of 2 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4945/storybook

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4945/storybook

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4945/storybook

packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
```diff
@@ -361,6 +365,18 @@ export class LuTooltipTriggerDirective implements OnDestroy {
 			]);
 	}
 
+	#resolveAnchor(): FlexibleConnectedPositionStrategyOrigin {
+		const anchor = this.luTooltipAnchor();
+
+		if (isNil(anchor)) {
+			return this.#host;
+		} else if ('elementRef' in anchor) {
```
Supamiu: I'd prefer to have a method to implement that, like a `getElementRef`, as using the field as-is means that it could use `elementRef` when, it's not needed.

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4945/storybook

Supamiu (approved):

PR #4949, opened 2026-06-03
fix(checkbox-input): presentation display not updated when control value changes

batnyu:
### 🐛 Problem

In presentation mode (`<lu-form-field presentation>`), when the control value changes programmatically (setValue(), patchValue(), reset…), the displayed value is not updated.

video: checkbox=true but the presentation display still shows "No"

https://github.com/user-attachments/assets/4c7e4b22-ffd3-4047-bfc6-b75384a4fd1b

### Root cause

The *luPresentationDisplayDefault templates interpolate ngControl.value, which is not reactive. The template is rendered through ngTemplateOutlet inside the OnPush FormFieldComponent as a transplanted view (declared in the input component, inserted into the form field): it only gets refreshed when the declaring input component's view is refreshed — and a setValue() marks neither of these views dirty.

Affected inputs: 
- checkbox, 
- switch, 
- number, 
- radio-group. 

The other ones only work by accident (text: ngx-mask side effect; textarea: a detectChanges() call for auto-resize; date/select: their writeValue updates internal signals).

### 🔧 Fix (scope: checkbox only)

Replaced the NoopValueAccessorDirective + injectNgControl() pattern with a proper ControlValueAccessor implementation:

- writeValue() feeds a value signal, read by the presentation template → the signal read makes the view reactive, even transplanted under OnPush
- native inputs are bound through [checked]="value()" / [disabled]="disabled()" / (change) / (blur) (both the regular branch and the filter pill branch)

video: presentation display now updates

https://github.com/user-attachments/assets/3db93f54-9af5-44c4-ba65-7f542e81d58d

If this approach is approved, follow-up PRs could cover number, switch and radio-group.

### ⚠️ Behavior changes

- The ng-invalid/ng-touched/… classes are no longer applied to the inner native input, only on the lu-checkbox-input host. No style in this repo targeted them (error state goes through [aria-invalid], set by the form field — unchang/after). Risk is limited to custom CSS / e2e selectors in consumer apps.

Note: an invalid + touched checkbox used without a form field still has no erring behavior (error styling relies on [aria-invalid], which only the form field sets; the old ng-* classes on the native input were never styled). Could be addressed for all inputs in a separate enhancement.

### ✅ Verified

- Presentation display updates on setValue() (after video)
- Edition mode: user click, ngModel, formControl, disabled, validation (aria-invalid, error message, rendering verified identical before/after, including markAllAsTouched() on a wrapping form group)
- Filter pill: interaction OK, QA page renders without errors
- Without form-field: bare / ngModel / formControl / disabled, including progr
- QA stories: new ValueUpdateBugRepro story (bug matrix across 8 input types in presentation mode), and filled the empty Angular cell of the Invalid row in the existing QA/Forms/Checkbox
page

-----

-----

### DIFF
--- a/packages/ng/forms/checkbox-input/checkbox-input.component.html
+++ b/packages/ng/forms/checkbox-input/checkbox-input.component.html
@@ -4,7 +4,10 @@
 		type="checkbox"
 		class="checkboxField-input"
 		[class.inputFramed-header-input]="framed"
-		[formControl]="ngControl.control"
+		[checked]="value()"
+		[disabled]="disabled()"
+		(change)="onCheckboxChange($event)"
+		(blur)="onTouched()"
 		[attr.aria-checked]="mixed() ? 'mixed' : null"
 	/>
 	<span class="checkboxField-icon" [class.inputFramed-header-icon]="framed" aria-hidden="true"
@@ -22,7 +25,10 @@
 			type="checkbox"
 			class="filterPill-checkbox-input"
 			luInput
-			[formControl]="ngControl.control"
+			[checked]="value()"
+			[disabled]="disabled()"
+			(change)="onCheckboxChange($event)"
+			(blur)="onTouched()"
 			[attr.aria-checked]="mixed() ? 'mixed' : null"
 			[attr.id]="filterPillInputId"
 		/>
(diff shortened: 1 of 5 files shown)


LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4949/storybook

Supamiu:
I feel like this should be closed because signal forms migration will fix it and presentation mode is not intended for realtime update in the first place, but will become compliant with it once signalForms are supported.

PR #4953, opened 2026-06-03
fix(empty-state-page): fix inline style issue with illustration

JeSuisCharlie1:
## Description

null in a binding like [style.--xxx] ="" not apply style which lets SCSS take advantage. To actually remove the background the ternary would need to return `none` instead of `null`.

-----

-----


fbasmaison-lucca (approved):

LuccaIntegration:
:woman_cook: https://lucca-front.lucca.tech/PR-4953/storybook

jeremie-lucca (approved):

jeremie-lucca:
Should be safe as these variables are only used with URL in prod

PR #5037, opened 2026-06-17
feat(rich-text-input): add secondary tag selector

obraud:
## Description

Add secondary tag selector

-----

-----

chrome_4SNUJmrP9L


packages/ng/forms/rich-text-input/plugins/tag/tag.component.html
```diff
@@ -16,4 +16,21 @@
...
 	}
+	@if (secondaryTags().length) {
+		<lu-filter-pill [label]="intl().selectTag">
+			<lu-simple-select
+				[formControl]="selectedTagControl"
+				[options]="filteredSecondaryTags()"
+				(clueChange)="tagSearch.set($event)"
+				#tagsSelect
+			>
+				@if (secondaryHasGroup()) {
```
obraud: Need a better solution to handle optional groups: as such, if group is not provided, an empty span is displayed in results. @jeremie-lucca @Supamiu ?

packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
```diff
@@ -127,4 +151,15 @@ export class RichTextPluginTagComponent implements RichTextPluginComponent, OnDe
...
+	groupByTagGroup(tag: Tag) {
+		return tag.group;
+	}
+
+	#normalize(str: string): string {
+		return str
+			.normalize('NFD')
+			.replace(/\p{Diacritic}/gu, '')
+			.toLowerCase();
+	}
```
obraud, line 174: Do we have a common utility file for this kind of code? @Supamiu

packages/ng/forms/rich-text-input/plugins/tag/tag.component.html
```diff
@@ -17,20 +17,36 @@
...
-				<ng-container *luOption="let tag; select: tagsSelect">{{ tag.description }}</ng-container>
-			</lu-simple-select>
-		</lu-filter-pill>
+		<button
+			luButton="ghost"
+			type="button"
+			size="XS"
+			[luPopover2]="tagsPopover"
+			[customPositions]="popoverPositions"
+			[attr.aria-expanded]="tagsTrigger.opened()"
```
obraud: Popover does not handle that natively?

packages/ng/forms/rich-text-input/plugins/tag/tag.component.html
```diff
@@ -17,20 +17,36 @@
...
+			<lu-icon icon="mathsPlus" />
+			{{ intl().selectTag }}
+		</button>
+
+		<ng-template #tagsPopover>
+			<div class="popover-contentOptional">
+				<lu-simple-select
+					[formControl]="selectedTagControl"
+					[options]="filteredSecondaryTags()"
+					(clueChange)="tagSearch.set($event)"
```
obraud, line 26:
```
(panelClosed)="closePopover()"
```

packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
```diff
@@ -89,8 +102,31 @@ export class RichTextPluginTagComponent implements RichTextPluginComponent, OnDe
...
 		});
+
+		// Render the select's panel inline inside our own popover (instead of through a filter-pill).
+		// The select component only appears once the popover opens, so we wire it up reactively.
+		effect(() => {
+			const ref = this.tagsSelectRef();
+			if (ref) {
+				untracked(() => {
+					ref.enableFilterPillMode();
+					ref.registerFilterPillClosePopover(this.closePopover);
```
obraud: If (panelClose) works, we should not need that line anymore

packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
```diff
@@ -89,8 +102,31 @@ export class RichTextPluginTagComponent implements RichTextPluginComponent, OnDe
...
+					ref.registerFilterPillUpdatePosition?.(this.updatePosition);
```
obraud: I guess the select panel is not exported by LF & cannot be used separately from select component? My concern with is that we register the callback but never unregister. Depending on internal implementation, that could lead to memory leaks & event handler duplications.

packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
MarcFairbrother: Looks like it works 🎉

Supamiu (dismissed):

Supamiu (dismissed):

jeremie-lucca (dismissed):

packages/ng/forms/rich-text-input/plugins/tag/tag.component.html
```diff
@@ -16,4 +16,27 @@
 			{{ tag.description }}
 		</lu-chip>
 	}
+	@if (secondaryTags().length) {
+		<lu-filter-pill [label]="intl().selectTag" class="tagSelect" #filterPillRef>
```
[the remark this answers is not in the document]
obraud: We now transfer focus back to the button.

packages/ng/forms/rich-text-input/plugins/tag/tag.component.ts
JeSuisCharlie1: No, we can create one if needed!

jeremie-lucca (approved):

PR #5067, opened 2026-06-23
tech(select): handle data source

JeSuisCharlie1:
## Description

Refactors the core select input to introduce a SelectDataSource<TOption> interface as the new first-class way to feed options into any select component. The SelectInputComponent exposes a dataSource model signal, and the API directive now builds and sets a SelectDataSource instead of directly subscribing to options.

**What changed**

- New SelectDataSource<TOption, TGroup> interface (select.model.ts) — encapsulates getOptions, getTotalCount, getGroupOptions, clueDebounceMs, paramsChange, and reset.

- New buildOptionsFromDataSource utility (select-input.utils.ts) — extracted, pure, observable pipeline for pagination + clue debounce + panel-open gating (with unit tests).

- ALuCoreSelectApiDirective — instead of imperatively pushing options, it now builds a SelectDataSource object and sets it on select.dataSource. The clue$ is now a BehaviorSubject (currentClue$). totalCount$ is now abstract and public. The TParams default changes from Record<string, ...> to Record<string, ...> | null.

- Simple/multi-select panels — updated to consume dataSourceOptions signal.

- Option comparer / key — calls on select.optionComparer()/select.optionKey() instead of direct property access.

**Beaking Changes**

- ALuCoreSelectApiDirective.totalCount$ — is now abstract and public. Subclasses must declare it (most already did; those that didn't will get a compile error).

- select.options is now an input() — programmatic assignment (select.options = [...]) is removed. Pass options via the [options] binding, or implement SelectDataSource and set select.dataSource

---------

---------

### DIFF
--- a/packages/ng/core-select/api/api.directive.ts
+++ b/packages/ng/core-select/api/api.directive.ts
@@ -1,6 +1,6 @@
 import { Directive, inject, OnDestroy, OnInit } from '@angular/core';
-import { ALuSelectInputComponent, coreSelectDefaultOptionComparer, coreSelectDefaultOptionKey, LuOptionComparer } from '@lucca-front/ng/core-select';
-import { catchError, combineLatest, concatMap, debounceTime, distinctUntilChanged, map, merge, Observable, of, pairwise, scan, startWith, Subject, switchMap, takeUntil, tap } from 'rxjs';
+import { ALuSelectInputComponent, coreSelectDefaultOptionComparer, coreSelectDefaultOptionKey, LuOptionComparer, SelectDataSource } from '@lucca-front/ng/core-select';
+import { catchError, combineLatest, concatMap, debounceTime, distinctUntilChanged, map, merge, Observable, of, pairwise, scan, startWith, Subject, switchMap, take, takeUntil, tap } from 'rxjs';
 
 export const LU_SELECT_MAGIC_PAGE_SIZE = 20;
 export const MAGIC_DEBOUNCE_DURATION = 250;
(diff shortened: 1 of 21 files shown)


packages/ng/core-select/api/api.directive.ts
```diff
@@ -40,6 +68,8 @@ export abstract class ALuCoreSelectApiDirective<TOption, TParams = Record<string
 	 */
 	protected abstract getOptions(params: TParams, page: number): Observable<TOption[]>;
 
+	readonly #lastPageByClue = new Map<string, number>();
```
GuillaumeNury: This Map will cause cache issues. Maybe storing lastClue and isLastPage in ngOnInit is enough

packages/ng/core-select/panel/panel.utils.ts
```diff
@@ -1,37 +1,40 @@
...
+		}
+
+		if (!searchable) {
+			return 'group-header';
+		}
+
+		return hasClue() ? 'option' : 'group-header';
+	});
+
+	// 		const getGroupTemplateLocation$ = (hasClue: boolean) =>
```
GuillaumeNury: Are we ok here? Can't remember

packages/ng/multi-select/panel/panel.component.ts
```diff
@@ -141,7 +131,20 @@ export class LuMultiSelectPanelComponent<T> implements AfterViewInit, CoreSelect
 		setTimeout(() => this.panelRef.updatePosition());
 	}
 
-	toggleOptions(notSelectedOptions: T[], groupOptions: T[]): void {
+	toggleOptions(notSelectedOptions: T[], groupOptions: T[], groupKey?: unknown): void {
+		const dataSource = this.selectInput.dataSource();
+		if (groupKey !== undefined && dataSource.getGroupOptions) {
+			this.groupLoadingKey.set(groupKey);
```
GuillaumeNury: We should handle multiple group loading (with a Set of loading groupkeys or a Map)

packages/ng/multi-select/panel/panel.component.ts
```diff
@@ -141,7 +131,20 @@ export class LuMultiSelectPanelComponent<T> implements AfterViewInit, CoreSelect
...
+			void firstValueFrom(dataSource.getGroupOptions(groupKey)).then((allGroupOptions) => {
+				this.groupLoadingKey.set(null);
```
GuillaumeNury: Should handle errors here. Stoping the loader is a good start

packages/ng/core-select/input/select-input.component.ts
```diff
@@ -356,6 +395,65 @@ export abstract class ALuSelectInputComponent<TOption, TValue> implements OnDest
 		}
 	}
 
+	#buildOptionsFromDataSource(ds: SelectDataSource<TOption>): Observable<readonly TOption[]> {
```
GuillaumeNury: I do not know if we can extract is beast somewhere else

packages/ng/core-select/panel/panel.utils.ts
JeSuisCharlie1: I’m going to run some tests on app's, but on my side it looks ok

jeremie-lucca (changes requested):
We lost the automatic panel repositioning with asynchronous options: 

https://github.com/user-attachments/assets/f23a0bf5-b289-4e8f-b59b-30467caf5a65

jeremie-lucca (dismissed):

jeremie-lucca (approved):

PR #5070, opened 2026-06-23
[activityFeed] add group for updates

### LINKED ISSUE #5031: [ActivityFeed] Multiple updates
### Intention
Refactor ActivityFeedUpdate to allow multiple updates

### Design
Image

### Key points
- We suggested to create a mandatory wrapper to ActivityFeedUpdate that will support the group UI (border, spacing, etc.)

vvalentin-lucca:
## Description

-----

close #5031 

-----

Capture d’écran 2026-06-23 à 17 24 53


packages/ng/activity-feed/activity-feed-group/activity-feed-group.component.ts
```diff
@@ -0,0 +1,14 @@
+import { NgTemplateOutlet } from '@angular/common';
+import { ChangeDetectionStrategy, Component, ViewEncapsulation } from '@angular/core';
+
+@Component({
+	selector: 'lu-activity-feed-group',
+	template: '<ng-content />',
+	encapsulation: ViewEncapsulation.None,
+	changeDetection: ChangeDetectionStrategy.OnPush,
+	imports: [NgTemplateOutlet],
```
Supamiu: This looks unused as the template doesn't use it.

packages/ng/activity-feed/activity-feed-update/activity-feed-update.component.html
```diff
@@ -1,11 +1,11 @@
-<div class="activityFeed-content-update-before">
-	<ng-container *ngTemplateOutlet="content" />
+<div class="activityFeed-content-group-update-object">
+	{{ label() }}
```
Supamiu: Using label like that here means that we no longer support `PortalContent`, we need to either fix that or keep it in mind as a breaking change.

vvalentin-lucca: This is an addition of a label, not a change to the "before" or "after."

jeremie-lucca:
Minor UI feedback
image
We should have: 
- 4px spacing between the label and the first value
- 6px spacing before and after the arrow

Figma provide a block including before / arrow / after to set the gap. It could be interesting to have the same structure for maintainability.

packages/ng/activity-feed/activity-feed-group/activity-feed-group.component.ts
```diff
@@ -0,0 +1,12 @@
+import { ChangeDetectionStrategy, Component, ViewEncapsulation } from '@angular/core';
+
+@Component({
+	selector: 'lu-activity-feed-group',
```
jeremie-lucca: I feel like it would be more precise is the name would contain the "update" information. 
As it's already breaking (with only one use), what about using this structure:
```
<lu-activity-feed-update>
  <lu-activity-feed-update-item>
  <lu-activity-feed-update-item>
</lu-activity-feed-update>
```

packages/scss/src/components/activityFeed/component.scss
```diff
@@ -65,18 +65,41 @@
...
+		.activityFeed-content-group-update-icon {
+			display: var(--components-activityFeed-content-group-update-icon-display);
 		}
 
-		.activityFeed-content-update {
-			display: var(--components-activityFeed-content-update-display);
+		.activityFeed-content-group {
+			padding: var(--pr-t-spacings-100) var(--pr-t-spacings-150);
+			border-radius: var(--pr-t-border-radius-structure);
+			border: 1px solid var(--palettes-neutral-300);
```
jeremie-lucca: We should use dedicated token : --pr-t-color-input-border

jeremie-lucca (changes requested):

vvalentin-lucca:
@jeremie-lucca The spacing between the colon and the text shouldn't be controlled by a spacing; it's text, and the spacing should be one space character.

jeremie-lucca (approved):

PR #5096, opened 2026-07-01
feat(core-select): add legal units directive with archived support

JeSuisCharlie1:
## Description

- Adds a new LegalUnitsDirective under `packages/ng/core-select/legal-units/` to fetch and display legal units via the API, following the same pattern as existing directives (users, departments, establishments).
- Introduces an ArchivedLegalUnitsComponent that displays a "Show archived" toggle inside the option panel.
- Wires the archived displayer into the mixin option item so consumers can opt into archived visibility.
- Adds i18n translations for the archived state label.
- Adds Storybook stories for both the basic and archived variants, for simple-select and the multi-select.

---------

---------

### DIFF
--- a/packages/ng/core-select/legal-units/archived-legal-units.component.ts
+++ b/packages/ng/core-select/legal-units/archived-legal-units.component.ts
@@ -1,0 +1,42 @@
+import { ChangeDetectionStrategy, Component, inject, InjectionToken, WritableSignal } from '@angular/core';
+import { outputToObservable, takeUntilDestroyed } from '@angular/core/rxjs-interop';
+import { FormsModule } from '@angular/forms';
+import { ɵCoreSelectPanelElement } from '@lucca-front/ng/core-select';
+
+export interface ArchivedLegalUnitsContext {
+	includeArchivedLegalUnits: WritableSignal<boolean>;
+}
+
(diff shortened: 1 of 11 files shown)


packages/ng/core-select/legal-units/archived-legal-units.component.ts
```diff
@@ -0,0 +1,45 @@
...
+	readonly context = inject(ARCHIVED_LEGAL_UNITS_CONTEXT);
+	readonly #selectableItem = inject(ɵCoreSelectPanelElement);
+
+	constructor() {
+		this.#selectableItem.id.set('select-archived-legal-units');
+		outputToObservable(this.#selectableItem.selected)
+			.pipe(takeUntilDestroyed())
+			.subscribe(() => {
+				this.context.includeArchivedLegalUnits.set(!this.context.includeArchivedLegalUnits());
+			});
```
Supamiu, line 37: Couldn't this be an effect now that we have them?

Supamiu (approved):

packages/ng/core-select/legal-units/archived-legal-units.component.ts
JeSuisCharlie1: selected is an output 😢 i could maybe do:
`const selected = toSignal(outputToObservable(this.#selectableItem.selected));`
and react to that signal but it doesn’t feel very cleany no?

jeremie-lucca (approved):

PR #5122, opened 2026-07-06
core(ng): enable strictNullChecks so published types model nullability

JeSuisCharlie1:
## Description

The library build inherited `strict: false` from the root `tsconfig.json`, so the
`.d.ts` files published to npm were generated in non-strict mode. Nullability was
not modeled in the public API, and consumers in strict mode inherited overly
optimistic types.

A parallel `tsconfig.lint.json` re-enabled `strictNullChecks: true` for the typed
lint pass only (via the `@lucca-front/ts-error` rule). So the `.ts` source was
already validated as strict-null clean, but the published artifact never benefited
from it — and the two configs could silently diverge.

## What

- **Enable `strictNullChecks: true` in the root `tsconfig.json`** (build level).
  The published `dist/ng` types now model nullability.
- **Remove the duplicate lint/build config**: delete `packages/ng/tsconfig.lint.json`
  and `.storybook/tsconfig.lint.json`, drop the redundant `strictNullChecks` override
  in `tsconfig.lint.json`, and repoint `eslint.config.mjs` at the base tsconfigs.
- **Fix template null-safety across ~30 entry points**. Enabling `strictNullChecks`
  reactivated Angular's `strictTemplates` null-narrowing — invisible to the lint pass
  (ESLint does not compile templates). Fixes are template-level and behavior-preserving:
  `@if (x(); as x)` narrowing, `?? undefined` / `?? null` coercions at binding sites,
  and optional chaining. **No public `input`/`output`/method signatures changed.**
- A handful of internal `.ts` accuracy fixes where no template-only fix was possible
  (e.g. `CalendarMonthInfo/YearInfo.rangeInfo` typed `RangeInfo | null` to match what
  `getRangeInfo()` actually returns, which let us delete two lying `as` casts). No
  `$any`, non-null assertions, or `@ts-ignore` were introduced. Also fixes a
  pre-existing `async` pipe precedence bug in `user-option` (`(x$ | async)`).
- **Stories stay out of the strict scope** (`strictNullChecks: false` in
  `.storybook/tsconfig.json`) — they are not published, consistent with the existing
  "stories can be strictified [later]" stance in `eslint.config.mjs`.

## Scope note

Only `strictNullChecks` is enabled; `strict` stays `false`. `noImplicitAny`,
`strictFunctionTypes`, `strictPropertyInitialization`, etc. are intentionally left
off — they are not validated by the current lint pass and would be a separate effort.

## Verification

- `npm run build:ng` — Built Angular Package (prisme + ng, all entry points) ✅
- `npm run lint:es` — 0 errors ✅
- `npm run build-storybook` — completed successfully ✅

-----

-----


packages/ng/callout/callout-icon.pipe.ts
```diff
@@ -12,7 +12,7 @@ import { getCalloutIcon } from './callout.utils';
 	name: 'luCalloutIcon',
 })
 export class CalloutIconPipe implements PipeTransform {
-	transform(state: CalloutState, icon: LuccaIcon): LuccaIcon | undefined {
+	transform(state: CalloutState | undefined, icon: LuccaIcon | undefined): LuccaIcon | undefined {
```
F2X, line 15: on ne devrait pas passer des non-nullables ici? ca n'a aucun sens de pipe des donnees nullish

Supamiu (dismissed):

F2X (dismissed):

packages/ng/callout/callout-icon.pipe.ts
JeSuisCharlie1: En soit on accepte le undefined sur le callout, on le gère comme ça aussi côté template

Supamiu (dismissed):

jeremie-lucca (approved):

PR #5123, opened 2026-07-06
refactor(file-upload): align single file upload with multi

JeSuisCharlie1:
## Description

Refactor lu-single-file-upload so it renders exactly like lu-multi-file-upload: the component is now just the dropzone, and the consumer owns the lu-file-entry and the toggle between the two.

Previously, SingleFileUploadComponent rendered the entry itself and exposed a set of pass-through inputs/outputs (entry, state, inlineMessageError, previewUrl, displayFileName, deleteFile). These are all removed. The single variant only ever holds one file, so once it's uploaded the consumer hides the dropzone and shows the lu-file-entry in its place — the consumer owns that toggle, the component does not do it anymore.

Changes
Remove the entry / state / inlineMessageError / previewUrl / displayFileName inputs and the deleteFile output from SingleFileUploadComponent.
Drop the entry rendering (@if (entry()) { <lu-file-entry ... /> }) and the is-hidden / disabled toggling from the template.
The component now reduces to the bare dropzone, keeping only the unnamed <ng-content /> (rendered in size="L") for in-dropzone projection (e.g. the AI tag) — same template shape as multi-file-upload.
Update the documentation and Storybook stories to the new sibling-entry usage (@if (fileUpload) { <lu-file-entry .../> } @else { <lu-single-file-upload .../> }).
Breaking change
Consumers must now render the lu-file-entry themselves as a sibling of lu-single-file-upload and toggle between the two, instead of binding [entry], [state], [previewUrl], [inlineMessageError], [displayFileName] and the (deleteFile) output.

Before

```
<lu-single-file-upload
  [accept]="accept"
  (filePicked)="uploadFiles([$event])"
  [entry]="fileUpload | luFileEntry"
  [state]="fileUpload?.state"
  [previewUrl]="getPreviewUrl(fileUpload)"
  [inlineMessageError]="fileUpload?.error?.detail"
  (deleteFile)="deleteFile(fileUpload)" />
```

After

```
@if (fileUpload) {
  <lu-file-entry
    [entry]="fileUpload | luFileEntry"
    [state]="fileUpload.state"
    [previewUrl]="getPreviewUrl(fileUpload)"
    [inlineMessageError]="fileUpload.error?.detail"
    (deleteFile)="deleteFile(fileUpload)" />
} @else {
  <lu-single-file-upload [accept]="accept" (filePicked)="uploadFiles([$event])" />
}
```

### DIFF
--- a/docs/file-upload.md
+++ b/docs/file-upload.md
@@ -33,21 +33,22 @@ Once the files are sent, the display management changes depending on the approac
 
 ### Single File Upload
 
-Just like the multi variant, you render the `lu-file-entry` yourself and project it into
-`lu-single-file-upload`. When a `lu-file-entry` is projected, the dropzone is hidden and replaced
-by the entry.
+Just like the multi variant, you render the `lu-file-entry` yourself, as a sibling of
+`lu-single-file-upload` rather than projected inside it. The single variant only ever holds one
+file, so once it's uploaded you hide the dropzone and show the entry in its place — you own that
+toggle, the component does not do it for you.
 
 ```html
 
 <lu-form-field label="Label">
   @let fileUpload = fileUploadFeature.fileUploads()[0];
-  <lu-single-file-upload [accept]="accept" (filePicked)="fileUploadFeature.uploadFiles([$event])">
-    @if (fileUpload) {
-      <lu-file-entry [entry]="fileUpload | luFileEntry" [state]="fileUpload.state"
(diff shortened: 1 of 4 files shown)


Supamiu:
The intended new usage would be to have the entry outside of the input so we can hide said input when the file is uploaded, I think it should be, at least documented, this way.

Supamiu (dismissed):

jeremie-lucca (approved):

PR #5124, opened 2026-07-06
fix(dialogHeader): update id with content

nbrun-lucca:
## Description

When the title/h1 changes in the dialog, then the aria-labelby is broken.

Example : passing this dialog from "readOnly" to "edit".

image

image

-----

### DIFF
--- a/packages/ng/dialog/dialog-header/dialog-header.component.ts
+++ b/packages/ng/dialog/dialog-header/dialog-header.component.ts
@@ -1,5 +1,5 @@
 import { CdkDialogContainer } from '@angular/cdk/dialog';
-import { ChangeDetectionStrategy, Component, contentChild, Directive, ElementRef, inject, input, OnInit, Renderer2, ViewEncapsulation } from '@angular/core';
+import { afterNextRender, ChangeDetectionStrategy, Component, contentChild, Directive, DoCheck, ElementRef, inject, Injector, input, OnDestroy, Renderer2, ViewEncapsulation } from '@angular/core';
 import { ButtonComponent } from '@lucca-front/ng/button';
 import { intlInputOptions } from '@lucca-front/ng/core';
 import { IconComponent } from '@lucca-front/ng/icon';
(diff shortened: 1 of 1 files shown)


packages/ng/dialog/dialog-header/dialog-header.component.ts
```diff
@@ -50,16 +54,39 @@ export class DialogHeaderComponent implements OnInit {
...
-		setTimeout(() => {
-			const header = this.#elementRef.nativeElement.querySelector('h1');
-			const id = header?.id || `lu-dialog-header-${nextId++}`;
-			if (header) {
-				this.#renderer.setAttribute(header, 'id', id);
-				this.#renderer.addClass(header, 'dialog-inside-header-container-title');
-			}
-			(this.#ref.cdkRef.containerInstance as CdkDialogContainer)?._addAriaLabelledBy(id);
-		});
+	ngDoCheck(): void {
```
shajz: ⚠️ [MEDIUM] The `ngDoCheck` + `afterNextRender` combo is going to run more than intended. `ngDoCheck` fires on *every* CD cycle for the whole lifetime of the dialog, and each pass schedules a fresh `afterNextRender` callback. The `id` guard keeps the expensive part (setAttribute/register) idempotent, but we still pay a `querySelector('h1')` after basically every render — it's polling in disguise.

Since the intent is "resync after render", `afterRenderEffect` expresses exactly that, registered **once**, without `DoCheck`, without the manual `Injector`, and without re-scheduling on each check:

```ts
constructor() {
  afterRenderEffect(() => this.#syncAriaLabelledBy());
}
```

You can then drop `implements DoCheck`, `ngDoCheck()`, and the `#injector` field. `ngOnDestroy` + the `#registeredAriaLabelledById` guard stay as-is.

(If profiling later shows resync-on-every-render is too costly, a `MutationObserver` scoped to the header container would only fire on actual DOM changes — but that's more moving parts, no need to preempt it.)

Supamiu (approved):

PR #5127, opened 2026-07-07
fix(navside): break long labels with hyphens

martincadoux:
## Description

Allow long module labels to be broken up with hyphens when necessary.

-----

**Before**
Capture d’écran 2026-07-07 à 15 26 18

**After**
Capture d’écran 2026-07-07 à 15 26 05

-----

### DIFF
--- a/packages/scss/src/components/navside/component.scss
+++ b/packages/scss/src/components/navside/component.scss
@@ -117,6 +117,7 @@
 		}
 
 		.navSide-item-link-title {
+			hyphens: auto;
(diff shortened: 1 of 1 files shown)


fbasmaison-lucca (approved):

JeSuisCharlie1 (approved):

jeremie-lucca:
As discussed with @vvalentin-lucca @martincadoux, it leads to extra word breaks on small words, that we'd like to avoid: 
image

PR #5128, opened 2026-07-07
fix(richText): prevent Shift+Enter soft line breaks inside lists

JeSuisCharlie1:
## Description

In the Rich Text editor, pressing Shift+Enter inside a list item used to insert a soft line break. This break could not be represented in Markdown, producing a broken/ambiguous list structure on export. Shift+Enter inside a list now creates a new list item (same as Enter), guaranteeing valid Markdown. Shift+Enter keeps its normal behaviour outside of lists.

-----

Added a `KEY_ENTER_COMMAND` handler in the list-format plugin (priority `NORMAL`, so it runs before `registerRichText`'s `EDITOR`-priority handler). It only acts when `shiftKey` is pressed **and** the selection is inside a `ListNode`; it then calls `preventDefault()` and redirects to `INSERT_PARAGRAPH_COMMAND`, consuming the event so no `LineBreakNode` is inserted. All other cases return `false` and fall through to default behaviour.

-----

### Contribution

- [ ] E2E test or UI diff is added if required
- [x] `npm run build` is OK
- [x] `npm run lint` is OK

### Functional review

- [ ] UI diff is OK
- [x] All related impacted functionalities are manually tested and show no regression

-----


obraud (approved):
👍

jeremie-lucca (approved):

PR #5129, opened 2026-07-07
FEAT vs code intellisense for lucca front

MarcFairbrother:
### Add CSS & SCSS IntelliSense for `@lucca-front/scss`

Editor tooling has always understood the Angular components but not the CSS side — custom properties (`--pr-t-*`, `--palettes-*`, `--commons-*`, `--breakpoints-*`) and `pr-u-*` utility classes are generated by SCSS loops, and the `commons/utils` mixins live in source files with no import barrel, so nothing could enumerate them for autocomplete. This PR adds a version-aware VS Code extension, backed by a manifest we now ship inside the library.

### What's in it

**1. CSS API manifest** (`feat(scss)`) — the build now compiles the full SCSS surface and emits `css-api/manifest.json` (617 custom properties, 1072 utility classes, 56 mixins) into the published `@lucca-front/scss` package. Each entry carries its value, category, and deprecation flag. In addition:

- **Resolved values** — utility declarations that reference custom properties also store the concrete resolved value (e.g. `var(--pr-t-spacings-100)` → `0.5rem`), so tooling can show both.
- **Validated replacements** — deprecated classes carry a recommended modern replacement, derived where possible (physical → logical, e.g. `pr-u-marginTop100` → `pr-u-marginBlockStart100`) and validated to point at an existing, non-deprecated class (chains through deprecated hops are followed; dead-ends are dropped).
- **Mixins** — the `@mixin` definitions in `commons/utils/*.scss` are parsed straight from source (they don't survive compilation to CSS) and emitted with their namespace, `@use` import path, full parameter signature, and any leading doc comment. Internal plumbing files (`index`, `namespace`, `highlight-prisme`) are excluded.

Component-scoped vars are excluded as internal. Deprecations come from `/* @deprecated */` source comments (stripped from the shipped compressed CSS, so consumers see no change) plus a small frozen seed for loop-generated names. Build-time self-checks guard against regressions (including mixin count and namespace/import sentinels). The `mixins` array is additive — the manifest stays at schema version 1, so it's forward- and backward-compatible.

**2. VS Code extension** (`feat(vscode-extension)`) — reads that manifest from the consumer's `node_modules`, so suggestions always match the installed version. Features:

- **Completion**
  - Custom properties in CSS/SCSS/LESS (including prefix-less inside `var(`, so `var(spacings` filters to spacing tokens), and `pr-u-*` classes in `class="…"` attributes in HTML and Angular inline templates. `--component-*` custom properties are purposefully excluded as they are intended for internal use.
  - **Mixins** (SCSS): after `@include `, suggests `namespace.mixin(…)` for every `commons/utils` mixin, with the parameter signature inline. Accepting a suggestion whose namespace isn't already imported **auto-inserts the matching `@use '@lucca-front/scss/src/commons/utils/<file>';`** at the top of the file's `@use` block.
- **Hover**
  - Custom properties: resolved value alongside the raw token chain.
  - Utility classes: the applied declarations, with the resolved value shown beneath any `var()`-based declaration.
  - Deprecated classes: surface the recommended replacement.
  - Unknown `pr-u-*` classes: suggest close matches (bounded edit-distance search over the known class names).
  - **Mixins**: `@include ns.mixin` shows the full signature, doc, and import path, and flags when the namespace isn't imported in the current file.
  - **Documentation links**: every token/class links to the Prisme design-system site; mapped utility families additionally deep-link to Storybook when a base URL is configured (see settings below).
- **Diagnostics** — unknown `pr-u-*` classes (catches typos and version mismatches, with a "did you mean…?" hint); **missing mixin imports** (SCSS): `@include ns.mixin` where `ns` is a known Lucca mixin namespace not `@use`d in the file — a genuine Sass compile error, so it's always on; and — when the experimental toggle is on — deprecated tokens/classes with their suggested replacement.
- **Quick Fixes** (`Ctrl+.` / `Cmd+.`) — replace an unknown class with a close match, replace a deprecated class with its recommended replacement, or **add the missing `@use` import** for an unimported mixin namespace.
- **Status bar** — a health snapshot of the open files (`LF <version> ⚠ <n>`), with a deprecated/unknown breakdown in the tooltip; click to open the Problems panel.

Zero runtime dependencies, compiled with `tsc`. Pure logic is isolated from the VS Code API and covered by 69 unit tests. CI builds and uploads the `.vsix` as an artifact.

### Settings

| Setting | Default | Description |
| --- | --- | --- |
| `luccaFront.diagnostics.enabled` | `true` | Toggle all diagnostics. |
| `luccaFront.diagnostics.deprecatedSeverity` | `warning` | Severity for deprecated tokens/classes. |
| `luccaFront.diagnostics.unknownClassSeverity` | `warning` | Severity for unknown `pr-u-*` classes. |
| `luccaFront.experimental.cssDeprecations` | `false` | **Experimental.** Flag deprecated custom properties and `pr-u-*` classes (see note below). |
| `luccaFront.docs.storybookBaseUrl` | `""` | Storybook base URL for per-family utility doc links. Fixed (`https://<host>/main/storybook`) or with a `{version}` placeholder filled from the installed version (`https://lucca-front.lucca.io/{version}/storybook` → `.../v21.0/storybook`). |
| `luccaFront.manifestPath` | `""` | Absolute path to a `manifest.json`, overriding node_modules discovery (for library development). |

### Notes

- **Deprecation flagging is experimental and off by default.** Deprecation signals (diagnostics, hover notices, completion strikethrough) are gated behind `luccaFront.experimental.cssDeprecations`, because deprecation metadata in `@lucca-front/scss` is still incomplete. Unknown-class detection and mixin import diagnostics are always on.
- **Manifest-only, forward-looking**: functional from the first release that ships the manifest onward. Older versions activate the extension but show a one-time notice and provide no suggestions (no crash, no wrong data). Manifests generated before mixin support simply offer no mixin features.
- Distributed as a **VSIX only** — not published to the Marketplace.

### Installing the extension

1. Download the `lucca-front-intellisense-*.vsix` artifact from this branch's CI run (or build locally: `npm run build:vscode-extension`).
2. Install it: `code --install-extension lucca-front-intellisense-0.4.0.vsix` (or in VS Code: Extensions panel → `⋯` → *Install from VSIX…*). If a previous version is already installed, uninstall it first — same-or-lower versions aren't replaced automatically.
3. Open a project with `@lucca-front/scss` installed. The status bar shows `LF <version>` once the manifest is loaded.

> Until a manifest-carrying version of `@lucca-front/scss` is released, point the `luccaFront.manifestPath` setting at a locally generated manifest (`npm run css-api` → `dist/scss/css-api/manifest.json`) to try it out.

### Testing done

Unit tests (69) + typecheck + lint pass; end-to-end run of the completion/hover/diagnostics pipeline against the real generated manifest (verifying resolved values, derived replacements, close-match suggestions, versioned doc links, and mixin extraction/import logic). The interactive Extension Development Host smoke test (F5 → *Run Lucca Front IntelliSense*) is the one manual step.

### Maintenance

The design keeps ongoing work low, but a few things need attention over time:

#### Routine / low effort
- **Deprecating a token or class**: add a `/* @deprecated <note> */` comment before the declaration in the SCSS source — the generator picks it up automatically. No generator or extension changes needed.
- **New tokens / utility classes / mixins**: nothing to do. New `commons/utils` mixins flow into the manifest on the next build (namespace, import path, and signature derived from source); tokens and classes do the same. All become available to consumers when they upgrade.

#### Occasional
- **Generator self-checks** (`packages/scss/css-api/generate.js`): the count thresholds and value/replacement/resolved-value/mixin sentinels are deliberately loose, but a large intentional change to the token/utility/mixin set may trip them — bump the numbers when that happens.
- **Mixin file scope**: the generator excludes `index`, `namespace`, and `highlight-prisme` from `commons/utils` as internal. If a new util file is added that should (or shouldn't) be consumer-facing, adjust `EXCLUDED_MIXIN_FILES`. Doc text comes from the leading source comment, so documenting a mixin is just a `//` comment above it.
- **Frozen deprecation seed** (`packages/scss/css-api/deprecations.js`): only for already-deprecated names that are generated by SCSS loops and can't carry a comment. It's a one-time snapshot; new deprecations should use the comment convention instead, not grow this list. This is also where physical → logical replacement derivations live — a new physical utility family would need a rule here (unresolved derivations are dropped and logged, so they fail safe).
- **Storybook doc-link map** (`packages/vscode-extension/src/docs/links.ts`): the class-prefix → family-slug table is hand-maintained. If a new documented utility family is added (or a story is renamed), add/adjust an entry; unmapped classes simply fall back to the Prisme link.
- **Extraction assumptions**: the generator relies on the config flags in `.storybook/styles.scss` (all palettes, deprecated prefix, cursive font) and on public custom-property prefixes. If the library adds a new public prefix or a new palette family, update `VARIABLE_CATEGORIES` / the config block in the generator.

#### Dependency / platform upkeep
- **`@vscode/vsce` and `@types/vscode`**: bump periodically. Raising `engines.vscode` (and the matching `@types/vscode` floor) unlocks newer VS Code APIs but drops support for older editors.
- **VS Code API changes**: the extension uses stable APIs only; watch for deprecations at major VS Code releases.

#### Known gaps that may warrant future work
- **Out-of-scope contexts** (documented in the README): `[class.x]` / `[ngClass]` bindings, `classList` in TS, and nested template literals inside `${…}`. Revisit if usage patterns shift.
- **Mixin import detection is namespace-based**: an unrelated `@use` that happens to expose the same namespace (e.g. a non-Lucca `media`) suppresses the missing-import diagnostic. This is a deliberate trade-off to avoid false positives; revisit if collisions prove common.
- **Deprecations without a modern twin**: some deprecated classes (e.g. deprecated font-size `text*` classes, `borderRadius` M/L/XL values) have no 1:1 replacement and are surfaced as deprecated with no suggestion — intended.
- **Tokens aren't deep-linkable**: token hovers link to the Prisme site root only, since neither Storybook nor ZeroHeight expose a per-token URL derivable from the token name.
- **No retro-compatibility**: versions of `@lucca-front/scss` published before the manifest existed get no suggestions. A CSS-parsing fallback (parsing the shipped `lucca-front.min.css`) could add basic support if backward reach becomes important.
- **Distribution**: VSIX-only means updates are manual. If adoption grows, consider Marketplace/internal-registry publishing (needs a publisher account and a release workflow).

-----

### LF version indicator

image

### Unsupported LF version

image

### Code completion for utility classes

Code_KPAgmTpwTe

### Code completion for custom properties

Code_k873tZTdcD

### Display information for utility classes

image

### Deprecated & Non existing

image

image

### Quick fix feature

Code_LFWMyHGy2d

### Code completion and quick fix imports for utility mixins

Code_LS56Iv6WDW

Code_Nt41Rez1tt


packages/vscode-extension/package.json
obraud, line 1: pas de package.lock ?

MarcFairbrother: Bonne question, je vais regarder pourquoi ça n'a pas été créé.

MarcFairbrother:
Replaced by https://github.com/LuccaSA/lucca-front/pull/5229 and https://github.com/LuccaSA/lucca-front/pull/5230

PR #5133, opened 2026-07-09
tech(ng): replace remaining setTimeout with afterNextRender and zoneless-friendly alternatives

JeSuisCharlie1:
## Description

Remove the last render/change-detection `setTimeout` calls from components, one of the final prerequisites for switching the project to zoneless.

Problem
When searching in an API-backed select (e.g. the establishment multi-select), typing a query, then selecting an option left the panel stuck on the filtered results. The search input was cleared, but the option list kept showing only the previously filtered items — the other options never came back until the panel was reopened.

Root cause
When an option is selected, the clue is immediately reset to '', which synchronously triggers getOptions({ clue: '' }) → buildParamsFromClue('').

However, the affected directives derive their request params from a signal:

protected readonly clue = toSignal(this.clue$);
protected readonly params$ = toObservable(computed(() => ({ ..., search: this.clue() ... })));
params$ (via toObservable) updates asynchronously. So buildParamsFromClue reading params$.pipe(take(1)) synchronously got the stale params, still containing search=Marseille. The API was therefore re-queried with the old search term, returning only the filtered results.

This is unlike api-v3, which builds params$ from a synchronous combineLatest([..., this.clue$]) and was not affected.

Fix
Override buildParamsFromClue(clue) in each affected directive to build the params directly from the clue argument instead of reading the async signal, guaranteeing fresh params on the immediate clue reset.

Affected directives: api-v4, establishments, users, legal-units, job-qualifications, occupation-category.

Test
Added EstablishmentTEST (Storybook play) covering the regression: open → search → select → assert the input is cleared and the full option list is restored (mouse + keyboard flows). Verified it fails on the pre-fix code and passes with the fix.

-----

Follow-up to #5092, covering the components it didn't touch:

- **`afterNextRender(..., { injector })`** for deferred DOM work: option scroll-into-view in `core-select` (drops the `MAGIC_OPTION_SCROLL_DELAY` constant, which wasn't part of the public API), the select-all header `scrollIntoView`, the deprecated `option-picker` highlight scroll, and the deprecated `lu-select` post-`ngAfterViewInit` `markForCheck`.
- **`core-select` `openPanel()`**: the `setTimeout` guarding against a double-tap is replaced by a synchronous `#isOpeningPanel` re-entrancy flag (the alternative the original author called out in the comment). Panel opening is now synchronous on click.
- **modal `dialog-content-adapter`**: the 2s error-class reset now uses `timer(2000)` + `takeUntilDestroyed`, so it's cancelled cleanly on destroy.
- **deprecated `option-picker`**: the init highlight reset now subscribes to the first `_options$` emission instead of `setTimeout(…, 1)`.

The genuine wall-clock timer in `humanize.utils.ts` (relative-time ticking) is left as-is — it's a real timer, not a rendering hack, and already emits through an Observable.

Specs importing `MAGIC_OPTION_SCROLL_DELAY` were updated to plain `tick()`.

-----

### DIFF
--- a/packages/ng/core-select/api/api-v4.directive.ts
+++ b/packages/ng/core-select/api/api-v4.directive.ts
@@ -3,7 +3,7 @@ import { computed, Directive, forwardRef, inject, input, model } from '@angular/
 import { toObservable, toSignal } from '@angular/core/rxjs-interop';
 import { ILuApiItem } from '@lucca-front/ng/api';
 import { applySearchDelimiter, CORE_SELECT_API_TOTAL_COUNT_PROVIDER, CoreSelectApiTotalCountProvider } from '@lucca-front/ng/core-select';
-import { debounceTime, map, Observable, switchMap } from 'rxjs';
+import { debounceTime, map, Observable, of, switchMap } from 'rxjs';
 import { ALuCoreSelectApiDirective } from './api.directive';
 
 @Directive({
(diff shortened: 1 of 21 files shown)


Supamiu (dismissed):

F2X (dismissed):

jeremie-lucca (dismissed):
Only tested with multi select, but it generates issues with provided data:

Before:

https://github.com/user-attachments/assets/f2e46470-9226-45cc-b62b-c2062f3bc8b4

After:


https://github.com/user-attachments/assets/b5086bba-c354-4f48-90af-8e340109b94d

Supamiu (dismissed):

vvalentin-lucca (dismissed):

Supamiu (approved):

vvalentin-lucca (approved):

PR #5149, opened 2026-07-16
feat: migrate select to use listbox component using Fable 5

Supamiu:
## Description

- [x] CSS file is not loaded by the Angular component
- [x] [Directives UI](https://pub-dc6d99acd6874e2aaff6219dd8a13ae2.r2.dev/PR-5149/index.html?path=/story/documentation-forms-simpleselect--establishment) is broken after a search (related to `listboxOption-content`'s display flex)
image

- [x] Groups are not migrated form `optionItem` in `ng/option/option.component.html`
- [x] Options update after search [with API](https://pub-dc6d99acd6874e2aaff6219dd8a13ae2.r2.dev/PR-5149/index.html?path=/story/documentation-forms-simpleselect--establishment)

https://github.com/user-attachments/assets/12f0b832-a055-4b5b-867b-1ed841f38d3a
- [x] Potential memory leak on [Tree](https://pub-dc6d99acd6874e2aaff6219dd8a13ae2.r2.dev/PR-5149/index.html?path=/story/documentation-forms-simpleselect--tree) after searching
- [x] [Department](https://pub-dc6d99acd6874e2aaff6219dd8a13ae2.r2.dev/PR-5149/index.html?path=/story/documentation-forms-multiselect--department) search is not filtering
image 

- [x] Checkbox option (select all, former employee, etc.) will not highlight on keyboard
- [x] Groups cannot be selected on keyboard

https://github.com/user-attachments/assets/a75b605e-4dc8-4c30-81af-086a9e3033e7

- [x] Multi select group options shouldn't be indented
image

- [x] All as default value : If I select a first option with keyboard, I can't interact with keyboard again except by selecting another option with mouse
- [x] Filter pills : Panel shadow is also visible inside the panel (+ double animation while opening)
image

- [x] Impersonation also have the double panel shadow issue
image

- [x] Filter pill : Can't close en pill by clicking on it 

https://github.com/user-attachments/assets/71440048-08a6-4664-aa05-a282320045a9

- [x] Spacing inconsistency between tree options 
image

- [x] Color picker layout is broken / no more zoom effect on highlight

https://github.com/user-attachments/assets/37923287-640d-4bec-b0ed-d911df6036ee

-----


JeSuisCharlie1 (dismissed):

F2X:
Not a blocker on this PR — a note so one hunk does not go through unnoticed, because it silently fixes a real defect on the base branch.

In `panel.component.ts`, this PR changes the post-load group toggle from `#applyGroupToggle(allGroupOptions, allGroupOptions)` to computing `notSelectedOptions` against the current selection. On `release/22.0` as it stands, that call is reached whenever `dataSource.getGroupOptions` exists — both call sites in `panel.component.html` always pass `group.key` — and it has two consequences:

- `enabledNotSelectedOptions` is never empty for a group with at least one enabled option, so the `else` branch ("all options already selected, unselect them all") is unreachable: the group toggle can select, never unselect.
- already-selected options are appended a second time to `selectedOptions`, which is then emitted as-is.

So the version of the hunk in this PR is the correct behaviour, not a refactor detail. Two things worth doing before merge:

1. Confirm the change is deliberate rather than incidental to the listbox migration.
2. Cover it with a spec. `panel.component.spec.ts` gains five listbox rendering tests here, but nothing exercises `getGroupOptions` — neither the loading path nor select-then-unselect on a group. That path is the whole point of #5067, and it is currently untested.

Context on why this matters beyond this PR: #5289. Read statically on `origin/release/22.0`, not reproduced at runtime.

<sub>— via Claude Code</sub>

jeremie-lucca (dismissed):

jeremie-lucca (approved):

shajz (commented):
Didn't have much time so take it or leave it

> 🤖 Review prepared with Claude — reviewed and edited by me.

packages/ng/core-select/panel/selectable-item.ts
```diff
@@ -8,11 +8,9 @@ import { CoreSelectPanelInstance, SELECT_PANEL_INSTANCE } from './panel.instance
 	selector: '[luCoreSelectPanelElement]',
 	exportAs: 'luCoreSelectPanelElement',
 	host: {
-		'[attr.id]': 'idAttribute()',
-		'[attr.aria-selected]': 'isSelected()',
 		'[class.is-highlighted]': 'isHighlighted()',
-		'(mouseenter)': 'onMouseEnter()',
-		role: 'option',
+		'(mouseover)': 'onMouseOver($event)',
+		role: 'presentation',
```
shajz, line 13: 🐞 **[MEDIUM]** `aria-activedescendant` points at ids that are no longer rendered

The host stopped binding `[attr.id]`, but the key manager still emits `idAttribute()` into `aria-activedescendant`. For the rows where the directive sits on the element itself instead of on a `lu-listbox-option`, nothing carries that id anymore: `lu-core-select-former-employees` (`select-former-employees`), `lu-core-select-archived-legal-units`, `lu-multi-select-all-header` (`multi-select-select-all`), and the per-group select-all `.link` (`<selectId>-group-<key>-action`). They highlight visually but a screen reader gets a dangling reference and announces nothing.

The `isSelectAllOption` helper added in `stories/helpers/test.ts` shows the same gap — its `option.id.includes('select-all')` branch can never match anymore.

Where the id should go is your call (back on those four hosts, or by wrapping them in a `lu-listbox-option`), so no suggestion here.

---

packages/ng/multi-select/panel/panel.component.html
```diff
@@ -1,121 +1,103 @@
...
-							(unselectMany)="toggleOptions([], $event)"
-						/>
-					}
-				} @else {
-					@for (option of options; track optionKey()(option); let index = $index) {
-						<lu-select-option
+		</ng-container>
+		@if (grouping() && groupTemplateLocation() === "group-header") {
+			@for (group of dataSourceOptions() | luOptionGroup: grouping().selector; track group.key; let groupIndex = $index) {
+				<lu-listbox-option group [groupId]="selectId + '-group-' + group.key" [elementId]="selectId + '-group-' + group.key">
```
shajz, line 20: 🐞 **[MEDIUM]** Group options render the same id twice and end up labelled by themselves

`[elementId]` puts `<selectId>-group-<key>` on the `lu-listbox-option` host, `[groupId]` puts the same value on the inner `.listboxOption-content`. So the id is in the document twice, and the host's `aria-labelledby` resolves to the first match — itself. The group's accessible name becomes its whole subtree, every option included.

Nothing reads that host id (the keyboard target is the separate `-action` element), so dropping it is enough. Same line in `simple-select/panel/panel.component.html:14`.

```suggestion
				<lu-listbox-option group [groupId]="selectId + '-group-' + group.key">
```

---

packages/ng/simple-select/panel/panel.component.html
```diff
@@ -1,90 +1,80 @@
...
-	</div>
-</div>
+		<ng-template #optionsList let-options let-groupIndex="groupIndex">
+			@if (treeGenerator) {
+				@for (branch of options | luTreeDisplay: treeGenerator; let index = $index; track optionKey()(branch.node)) {
+					<lu-tree-branch
+						[branch]="branch"
+						[optionTpl]="optionTpl()"
+						[optionComparer]="optionComparer()"
+						[selectedOptions]="[selected()]"
```
shajz, line 35: 🐞 **[MEDIUM]** `[selected()]` hands `[null]` to the tree branches while nothing is selected

`selected()` is `null` until a value is picked, so `luIsOptionSelected` runs the comparer against `null`. `LuCoreSelectDepartmentsDirective.optionKey` guards for that; `api-v4`'s `optionKey = (option) => option.id` does not, so a tree select built on the v4 directive throws as soon as the panel opens.

The `selectedOptions` computed declared right above returns `[]` in that case, and its comment says it exists for exactly this — it is just no longer used.

```suggestion
						[selectedOptions]="selectedOptions()"
```

---

packages/scss/src/components/listboxOption/index.scss
```diff
@@ -93,9 +121,16 @@
 	}
 
 	@layer mods {
-		&.is-hovered,
 		&:hover {
-			@include hover;
+			.selectPanel.is-pointerNavigation & {
```
shajz, line 125: ⚠️ **[MEDIUM]** The `scss` sheet now depends on `.selectPanel`, a class only the Angular panels emit

`.selectPanel` comes from the `host` metadata of the two select panels, so outside a select none of these hover rules match — neither the `:hover` branch nor the `.is-hovered` one. The standalone listbox documented under `Documentation/Forms/Listbox Option/HTML&CSS` loses its hover feedback on single-select options and on group headers (the `mod-multiple` block is not gated, so multiple keeps it).

It also inverts the published dependency order: `ng` depends on `scss`, not the reverse. Gating on a class this sheet owns, that the panel sets, would keep the pointer/keyboard split without pulling an Angular-side name into the design system.

---

packages/scss/src/components/listboxOption/vars.scss
```diff
@@ -23,10 +23,10 @@
 	--components-listboxOption-content-position: relative;
 	--components-listboxOption-content-before-content: var(--components-listboxOptionWrapper-childrenBefore-content, '');
 	--components-listboxOption-content-after-content: none;
-	--components-listboxOption-content-opacity: 1;
 	--components-listboxOption-content-boxShadow: none;
 	--components-listboxOption-content-flexWrap: nowrap;
 	--components-listboxOption-content-justifyContent: normal;
+	--components-listboxOption-content-link-opacity: 0;
```
shajz, line 29: ⚠️ **[MEDIUM]** The link-opacity default hides every link inside an option, not just the group select-all

`varsContent` runs on every `.listboxOption-content`, and `component.scss` now applies `opacity: var(--components-listboxOption-content-link-opacity)` to any `.link` inside it. A consumer putting a link in an option template gets an invisible — but still focusable and clickable — element.

The previous version scoped the trick to the group title (`.lu-picker-content-group:not(:hover, :has(.is-highlighted)) .link`). Defaulting this var to `1` and setting `0` inside `@mixin group` would keep the reveal-on-hover where it is wanted.

---

packages/ng/core-select/establishment/establishment-grouping.service.ts
```diff
@@ -13,7 +13,7 @@ export class EstablishmentGroupingService {
 	protected readonly legalUnitsCount$ = this.http.get<{ count: number }>(this.legalUnitsUrl, { params: this.countParams }).pipe(map((res) => res.count));
 
 	public readonly useGrouping$ = combineLatest([this.legalUnitsCount$, this.establishmentsCount$]).pipe(
-		map(([luCount, establishmentCount]) => luCount > 1 && establishmentCount > 1 && luCount !== establishmentCount),
+		map(([luCount, establishmentCount]) => luCount > 1 && establishmentCount > 1),
```
shajz, line 16: ⚠️ **[MEDIUM]** Dropping `luCount !== establishmentCount` changes what every establishment select shows

Grouping now kicks in for tenants where each legal unit holds exactly one establishment — the case where the group header just repeats the option under it, which is what the condition was there to avoid. The commit calls the condition arbitrary, but the PR description doesn't list this among the fixes and no spec covers it.

Since it ships to every product using an establishment select, it is worth a line in the description and in `core-select/CHANGELOG.md`.

---

packages/ng/core-select/option/option.component.scss
```diff
@@ -2,39 +2,50 @@
...
 		}
 	}
 
-	.optionItem-value:has(.optionItem-icon) {
-		display: flex;
-		justify-content: space-between;
+	.optionItem-value {
+		padding-block: var(--components-options-item-padding-vertical);
+		padding-inline: var(--components-options-item-padding-horizontal);
 	}
```
shajz, line 25: 🧹 **[LOW]** These `.optionItem` / `.optionItem-value` rules no longer match anything

The host lost its `optionItem` class and the `.optionItem-value` wrapper gave way to `lu-listbox-option`, so under emulated encapsulation neither selector has an element left in this template — `simple-select/panel/panel.component.spec.ts` even asserts `.optionItem-value` is null. `.optionItem-icons` and below still apply, and the hover reveal now comes from `tree-branch.component.scss`. Non-blocking.

```suggestion
@layer components {
```

---

packages/ng/listbox/option/option.component.html
```diff
@@ -9,8 +9,9 @@
 		<lu-icon icon="mathsPlus" />
 	}
 	<ng-content />
+	<ng-content select="[group-action]" />
 	@if (selectAll()) {
-		<lu-listbox-option select>{{ selectAll() }}</lu-listbox-option>
+		<lu-listbox-option [elementId]="elementId() + '-selectAll'" select>{{ selectAll() }}</lu-listbox-option>
```
shajz, line 14: 🧹 **[LOW]** `elementId()` defaults to `null`, so this renders `id="null-selectAll"`

`elementId` is `input<string | null>(null)` and no caller sets it on a `selectAll` option today (the group story uses the bare `selectAll="…"` form), so the nested option ships a literal `null-selectAll` id. Non-blocking.

```suggestion
		<lu-listbox-option [elementId]="elementId() ? elementId() + '-selectAll' : null" select>{{ selectAll() }}</lu-listbox-option>
```

---

PR #5157, opened 2026-07-17
fix(grid): make --grid-maxWidth a true maximum instead of a fixed width

### LINKED ISSUE #5188: [Grid] mode="form" inside a .form collapses — inline-size resolves to the invalid `none`
## Description

`grid/component.scss` consumes `--grid-maxWidth` as a hard width rather than a maximum, with the `max-inline-size` variant sitting right above it, commented out:

```scss
// max-inline-size: var(--grid-max-width, var(--grid-maxWidth));
inline-size: var(--grid-max-width, var(--grid-maxWidth));
max-inline-size: 100%;
```

Since #4727, `grid/mods.scss` feeds that variable from the form one:

```scss
@mixin form {
	--grid-columns: 4;
	--grid-maxWidth: var(--components-form-maxWidth, 50rem);
}
```

and `form/vars.scss` declares `--components-form-maxWidth: none`.

Those vars are emitted on `.form` (not on `:root`), so any `lu-grid mode="form"` nested in a `<form luForm>` inherits `none` and ends up with `inline-size: none` — invalid for that property. The declaration is dropped at computed-value time and the grid falls back to `inline-size: auto`.

The `50rem` fallback never kicks in: the variable *is* defined, just to a value that is invalid for `inline-size`.

## Impact

In a stretch context this is invisible. In a shrink-to-fit context (flex/grid item aligned to `start`, float, absolute positioning) the grid collapses to its max-content width instead of filling its container.

Before 21.3, `--components-form-maxWidth` was `40rem`, so the grid resolved to `inline-size: 40rem` and the problem was masked.

## Reproduction

A `lu-grid mode="form"` inside a `<form luForm>`, itself a child of a flex column whose cross-axis alignment is `start`:

```html
<form luForm>
  <lu-grid mode="form">
    <lu-grid-column colspan="4">…</lu-grid-column>
  </lu-grid>
</form>
```

21.2.4 → grid is 40rem wide. 21.3.0 and later → grid shrink-wraps its content.

## Suggested fix

Use the commented-out declaration, which is what the variable name promises:

```scss
max-inline-size: var(--grid-max-width, var(--grid-maxWidth));
```

`max-inline-size: none` is valid, so `none` then behaves as the no-op #4727 intended.

## Affected versions

21.3.0 through current `master`.

`fieldset/vars.scss` reads the same variable (`--components-fieldset-contentMaxWidth: var(--components-form-maxWidth, 50rem)`) and may deserve the same check.

fbasmaison-lucca:
## Description

* Grid: `--grid-maxWidth` now acts as a true maximum, not a fixed width.
* Form grids keep an auto inline-size.
* They no longer occupy 50rem when their content is narrower.

---

## Note for reviewers

* [A few occurrences](<https://github.com/search?q=org%3ALuccaSA+%2F%28--grid-maxWidth%7C--grid-max-width%29%2F+NOT+repo%3ALuccaSA%2Flucca-front+&type=code>) of `--grid-max-width` found outside LF.
* [#3154](<https://github.com/LuccaSA/lucca-front/issues/3154>) reverted this same approach.
  * It broke horizontal fieldsets in their flex-based layout at the time.
* The fieldset now uses a CSS grid layout (`grid-template-columns: 1fr auto`).
  * A child grid stretches to fill its cell.
  * The old breakage should no longer apply.

## Context

* `mod-form` sets `--grid-maxWidth: var(--components-form-maxWidth, 50rem)`.
* The grid consumed it through `inline-size`.
  * `max-inline-size: 100%` clamped it.
  * Result: a form grid always occupied 50rem, even with narrower content.
* The default `--grid-maxWidth: none` is invalid for `inline-size`.
  * It resolved to `auto` only through IACVT.
  * It silently overrode any other `inline-size` declaration in the cascade.
* Downstream symptom example: spurious horizontal scrollbar in a dialog.
  * The form grid forced its fieldset wider than the dialog content.
  * Sidenote: screenshots below come from LuccaSA/Cleemy.Front#905.
  * That PR works around the issue on the product side with `pr-u-minInlineSize0`.

Before:

Card order dialog with spurious horizontal scrollbar

After:

Card order dialog without horizontal scrollbar

## Solution

* Apply the variable through `max-inline-size` instead of `inline-size`.
  * This matches the original intent (commented-out line).
* Remove the `max-inline-size: 100%` clamp.
  * Nothing can force an overflow without a fixed inline-size.
* The default `none` is a valid value for `max-inline-size`.
  * No more reliance on IACVT.
* Align the grid with the form component.
  * The form already uses `max-inline-size: var(--components-form-maxWidth)`.

---

Fixes #5188

---


shajz:
@fbasmaison-lucca 
Not a reviewer anymore but I notice this PR fixes my issue [#5188](https://github.com/LuccaSA/lucca-front/issues/5188) (`inline-size: none` invalid, resolved as `auto` by IACVT). 

Might you add a tiny `Fixes #5188` in the description please ? 😊

vvalentin-lucca:
This breaks the horizontal layout of the `fieldset` component.

Before: 

Capture d’écran 2026-08-17 à 12 02 56

After: 

Capture d’écran 2026-08-17 à 12 03 21

jeremie-lucca:
Will not be merged as we can't find a good option for now. 
Issue (breaking) targeting 23.0 https://github.com/LuccaSA/lucca-front/issues/5333

PR #5172, opened 2026-07-22
fix(core-select): start the loader as soon as the panel opens

batnyu:
## Description

Fixes a flash of the "no value available" empty state when opening any API-backed select (`lu-simple-select`/`lu-multi-select` with `users`, `departments`, `establishments`, api v3/v4 directives…).

### Why it happened

The panel shows its empty state whenever `options.length === 0 && !loading`. On opening, the panel is rendered in the same task as the `isPanelOpen$` emission, but the fetch pipeline only reacts one or more tasks later (`debounceTime(0)` on the open stream, async `params$` — e.g. a `toObservable` in the users directive). `loading` was only set to `true` inside `getOptionsPage()`, i.e. in that later task: the panel's first paint happened with `loading === false`, showing the empty state until the fetch started.

The window is made of task scheduling, so its duration scales with CPU load: barely a frame on an idle machine, but clearly visible on a busy dev server (measured ~130ms at 6x CPU throttling, ~1.5s at 60x).

There was already a guard for this ("Start the loader as soon as the panel is opened to avoid a short display of the 'no result' message"), but it only covered opening with a pending clue debounce, and sat *after* the `debounceTime(0)` — one task too late for the first paint.

### The fix

Start the loader synchronously with the `isPanelOpen$` emission (same task as the panel creation), unconditionally: every open leads to a fetch, which always resolves the loader (response or `catchError`). The pending-clue case is a subset of the new behavior, and the anti-double-call logic downstream is untouched (see the existing search specs, still green).

A regression test asserts that `loading` is already `true` at the moment `isPanelOpen$` delivers the open event — the exact condition that decides what the panel's first paint shows. It fails on the previous implementation.

## Videos

Recorded on the `SimpleSelect / User Select` story, **unmodified code**, with devtools CPU throttling **x60** to widen the window (the same flash occurs at every speed, just shorter):

### Before (empty state flashes on open)

[core-select-before.webm](https://github.com/user-attachments/assets/83581c61-127d-48d7-95af-e91015143267)

### After (loader from the very first paint)

[core-select-after.webm](https://github.com/user-attachments/assets/e9fb8ae8-2228-4b40-95fd-8b48592dd318)

-----

### DIFF
--- a/packages/ng/core-select/api/api.directive.ts
+++ b/packages/ng/core-select/api/api.directive.ts
@@ -57,16 +57,21 @@ export abstract class ALuCoreSelectApiDirective<TOption, TParams = Record<string
 	protected buildOptions(): Observable<TOption[]> {
 		// Prevent a double call to getOptions when the clue is changed while the panel is closed
 		const clueIsPendingDebounce$ = merge(this.select.clueChange$.pipe(map(() => true)), this.clue$.pipe(map(() => false))).pipe(distinctUntilChanged());
-		const isOpen$ = combineLatest([this.select.isPanelOpen$, clueIsPendingDebounce$]).pipe(
+		const isOpen$ = combineLatest([
+			this.select.isPanelOpen$.pipe(
+				tap((isOpen) => {
+					// Start the loader synchronously on opening to avoid a short display of the "no result"
+					// message: the first fetch only starts on the next tick (debounced open, async params$)
+					if (isOpen) {
+						this.select.loading = true;
+					}
+				}),
+			),
+			clueIsPendingDebounce$,
+		]).pipe(
 			debounceTime(0),
 			startWith([false, false]),
 			pairwise(),
-			tap(([[wasOpen], [isOpen, clueIsPendingDebounce]]) => {
-				// Start the loader as soon as the panel is opened to avoid a short display of the "no result" message
-				if (!wasOpen && isOpen && clueIsPendingDebounce) {
-					this.select.loading = true;
-				}
-			}),
 			map(([[wasOpen], [isOpen, clueIsPendingDebounce]]) => (isOpen && !wasOpen ? !clueIsPendingDebounce : isOpen)),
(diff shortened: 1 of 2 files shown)


packages/ng/core-select/api/api.directive.ts
```diff
@@ -57,16 +57,21 @@ export abstract class ALuCoreSelectApiDirective<TOption, TParams = Record<string
...
 		// Prevent a double call to getOptions when the clue is changed while the panel is closed
 		const clueIsPendingDebounce$ = merge(this.select.clueChange$.pipe(map(() => true)), this.clue$.pipe(map(() => false))).pipe(distinctUntilChanged());
-		const isOpen$ = combineLatest([this.select.isPanelOpen$, clueIsPendingDebounce$]).pipe(
+		const isOpen$ = combineLatest([
+			this.select.isPanelOpen$.pipe(
+				tap((isOpen) => {
+					// Start the loader synchronously on opening to avoid a short display of the "no result"
+					// message: the first fetch only starts on the next tick (debounced open, async params$)
+					if (isOpen) {
+						this.select.loading = true;
```
Supamiu, line 66: This will show loader everytime we open it instead of only on the first occurence, I think that's what the previous approach did.

batnyu: I think it's already the case. In the "user" tab of this link https://prisme.lucca.io/94310e217/p/587833-select-simple, everytime we open it, it shows the spinner

JeSuisCharlie1 (approved):

packages/ng/core-select/api/api.directive.ts
jeremie-lucca: As our select flush the options after closing the panel if there is an API call, I guess it doesn't have any negative impact? I cannot see any side effect while testing.

Supamiu (approved):

PR #5095, opened 2026-07-01
chore(stylelint): Upgrade to 17.14.1 with fixes

fbasmaison-lucca:
## Description

Upgrade `@lucca-front/stylelint-config` to Stylelint 17.14.1 and prepare the enforcement of logical, RTL-ready CSS.

- New rule: [`selector-no-deprecated`](https://stylelint.io/user-guide/rules/selector-no-deprecated/) catches deprecated selectors.
- The logical-properties rules ship with the SCSS migration in the stacked follow-up PR #5207.

> [!NOTE]
> See [first comment](https://github.com/LuccaSA/lucca-front/pull/5095#issuecomment-4879234772) about undocumented deprecations.

-----

Technical notes for reviewers.

### Dependencies

- Bump peers:
  - `stylelint ^17.14.1`;
  - `@stylistic/stylelint-plugin ^5.2.1`;
  - `@stylistic/stylelint-config ^5.0.0`.

### Hardening (`stylelintForLF.mjs`)

- Compare LF versions numerically; order prereleases before their release.
- Derive message status and severity from a single source.
- Paginate the milestones API and store dates as ISO.
- Ignore disallowed objects with a missing `objectPattern` instead of invalidating the whole rule.

### Deprecation data

- Add exact-match boundaries to every pattern.
- Dedupe overlapping selector entries; link a regex101 test to each.
- Repair the dead elevation rule; restrict palettes to the real shades.

### Other

- Add `languageOptions.directionality` (block `top-to-bottom`, inline `left-to-right`).
- Disable `@stylistic/no-multiple-whitespaces`.
- Unify lint scripts under the `lint` namespace.

-----

PR #4884, opened 2026-05-21
[calendar] week number

vvalentin-lucca:
## Description

HTML version of the component with week selection.

-----

⚠ Update doc to note that the returned date is Thursday of the selected week, at midnight local time.

-----

Capture d’écran 2026-05-29 à 15 29 23

PR #5206, opened 2026-07-31
fix(calendar2): Type the week-day cell tooltip under strictNullChecks

fbasmaison-lucca:
## Description

Restore the `release/22.0` build: the calendar week view's day tooltip no longer breaks compilation under `strictNullChecks`.

-----

### Context

- #5122 typed the month-view day tooltip with a `?? ''` fallback.
- #4884 added the week-number row with a second, untyped `[luTooltip]="day.label"`.
- The two PRs merged in close succession; their combination does not compile:
	- `day.label` is `string | undefined`.
	- `luTooltip` expects `string | SafeHtml`.
- Every `release/22.0` build fails since, and PR checks inherit the breakage (e.g. #5095).

### Fix

Apply the same `?? ''` fallback to the week-row binding.

-----

### Contribution

- [ ] E2E test or UI diff is added if required
- [x] `npm run build` is OK
- [x] `npm run lint` is OK

### Functional review

- [ ] UI diff is OK

-----

### DIFF
--- a/packages/ng/date2/calendar2/calendar2.component.html
+++ b/packages/ng/date2/calendar2/calendar2.component.html
@@ -160,7 +160,7 @@
 											(keydown.space)="day.disabled ? null : onCellClicked(day.date)"
 											(focus)="day.isOverflow ? null : dateHovered.set(day.date)"
 											(blur)="day.isOverflow ? null : dateHovered.set(null)"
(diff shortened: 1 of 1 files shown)


fbasmaison-lucca:
Error disappeared after pulling updated release/22.0 branch.

PR #4270, opened 2025-12-10
chore: remove StoryFn usages

GuillaumeNury:
## Description

Functional description for the changelog.
Keep an empty line under your text, as well as the 5 lines that follow it.

-----

⚠️ Please don't merge before 20.3.4 release ⚠️

-----

PR #5208, opened 2026-08-03
fix(core-select): expose aria-disabled on disabled options

shajz:
## Description

Screen readers announce an option disabled with `luDisabledOption` as selectable, when it isn't.

`luDisabledOption` pushes into the option context, which `LuOptionComponent` applies as `is-disabled` on the inner `.optionItem-value`: the option greys out and swallows its click. The `[role="option"]` host, carried by `CoreSelectPanelElement`, never reflects it — so the visual and functional blocks are there, only the accessible information is missing.

`CoreSelectPanelElement` already receives that state as its `disabled` input, which it uses to ignore `mouseenter`. It now also exposes it as `aria-disabled`.

<details><summary>Notes for the reviewer</summary>

`disabled` is a plain `@Input`, mutated imperatively from the subscription in `LuOptionComponent.ngAfterViewInit`, not a signal. That was the only real risk of the change: the host binding still updates, because the `markForCheck()` that already sits next to the assignment marks the declaring view dirty. Both tests below fail without the one-line fix, which is what makes them worth reading.

`'disabled || null'` keeps `aria-disabled` off the enabled options rather than spraying `aria-disabled="false"` over the whole panel.

Two levels of coverage, since there was none on `luDisabledOption`:

- a unit spec on `CoreSelectPanelElement`, asserting the attribute lands on the right option and stays off the others;
- `WithDisabledOptionsTEST`, restored. It was the only commented-out test story in the repo, against 61 active ones, and it went down in the 488-file chore of #4270 rather than by decision. It asserted the `is-disabled` class on the inner element; it now also asserts `aria-disabled` on the option itself.

Only the simple-select story is covered: the fix sits in `CoreSelectPanelElement`, shared by both, and the multi-select story had no test story to restore.

Keyboard navigation still skips disabled options entirely, which the ARIA APG argues against for listbox options — that one is a behaviour change, out of scope here.

</details>

-----

### DIFF
--- a/packages/ng/core-select/panel/selectable-item.spec.ts
+++ b/packages/ng/core-select/panel/selectable-item.spec.ts
@@ -1,0 +1,57 @@
+import { ChangeDetectionStrategy, Component } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { FormsModule } from '@angular/forms';
+import { LuDisabledOptionDirective, LuOptionDirective } from '@lucca-front/ng/core-select';
+import { LuSimpleSelectInputComponent } from '@lucca-front/ng/simple-select';
+import { vi } from 'vitest';
+
+type Entity = { id: number; name: string };
+
+const options: Entity[] = [
+	{ id: 1, name: 'test 1' },
+	{ id: 2, name: 'test 2' },
+];
+
+@Component({
+	selector: 'lu-disabled-option-host',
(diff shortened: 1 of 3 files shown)


Supamiu:
Looks like tests aren't passing, could you check them and fix accordingly please? 🙂

jeremie-lucca:
Fixed by https://github.com/LuccaSA/lucca-front/pull/5149

PR #5221, opened 2026-08-04
fix(core-select): keep the user select loader on until the "me" lookup resolves

batnyu:
## Description

Fixes two failure modes of the user selects (`lu-simple-select`/`lu-multi-select` with the `users` directive) around the "me" lookup request:

- the "no value available" empty state flashed on opening, for the duration between the users search response and the "me" lookup response;
- a failed "me" lookup left the select stuck forever (no options, previously a permanent empty state), since the options stream never emitted.

-----

### Why it happened

`getOptionsPage()` runs two parallel requests on an empty clue: the users search and the "me" lookup (the option prepended on the first page). The loader was turned off by a `tap` on the **users search response**, but the options only come out of the `combineLatest([me$, users$])` — so between the two responses the panel had `options=[] && !loading`, which is exactly its empty-state condition. Locally the window is one frame; it scales with the latency gap between the two requests (measured 2.5s of empty state with a 2.5s-delayed "me" lookup, see videos).

This is the remaining window after #5172: that fix guarantees the loader is on when the panel opens (the entry of the cycle), this one keeps it on until the options actually arrive (the exit).

On top of that, `me$` swallowed errors with `catchError(() => EMPTY)`: an EMPTY completes without emitting, so a failed "me" lookup froze the `combineLatest` forever. It now emits `null` — the select degrades gracefully by displaying the options without the "me" entry. This is a deliberate behavior change: the "me" option is best-effort, its failure should not discard a successful users search (nor display the misleading "no value available, contact your administrator" message).

Each regression test fails on the previous implementation for its own fix only (verified by applying each fix in isolation).

-----

### Contribution

- [ ] `npm run build` is OK
- [ ] `npm run lint` is OK

### Functional review

- [ ] UI diff is OK

-----

### DIFF
--- a/packages/ng/core-select/user/users.directive.spec.ts
+++ b/packages/ng/core-select/user/users.directive.spec.ts
@@ -159,6 +159,64 @@ describe('LuCoreSelectUsersDirective', () => {
 		httpTestingController.verify();
 	}));
 
+	it('should keep the loader on until the "me" lookup answers', fakeAsync(() => {
+		// Arrange
+		simpleSelect.openPanel();
+		fixture.detectChanges();
+		tick(MAGIC_OPTION_SCROLL_DELAY);
+
+		const meUser = createUser(CURRENT_USER_ID);
+		const users = [createUser(1), createUser(2)];
+
+		const meReq = httpTestingController.expectOne(`/api/v3/users/search?fields=${fields}&id=${CURRENT_USER_ID}`);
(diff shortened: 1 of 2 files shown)


lmullot (approved):
LGFM

F2X (approved):

Supamiu (approved):

PR #5228, opened 2026-08-07
fix(user-popover): uncaugth exception on API errors

obraud:
## Description

Fixed uncaugth exception when user-popover API call fails.

-----

User-popover image caching operator did not handle exceptions coming from source observable. It broke the rest of the pipe and ended in console.
* Unwanted logs in datadog
* Infinite skeleton
* Fallback behavior with basic info was never visible

-----
Before:
firefox_p5rX2GKjr1
After: 
firefox_OBa6v4OQWB


obraud:
@shajz That's the source of most logs on Home page

shajz (approved):
🤯

lmullot (approved):

Supamiu (approved):

PR #5229, opened 2026-08-07
feat(scss): ship a CSS API manifest and a Storybook explorer

MarcFairbrother:
## Description

`@lucca-front/scss` now ships a machine-readable description of its CSS API, plus a Storybook page to browse it.

- **`css-api/manifest.json`** — 617 custom properties, 1082 `pr-u-*` utilities, 56 mixins
- Generated from the compiled CSS, so it cannot drift from what the library ships
- Carries resolved values, deprecation status, and replacements where they exist
- **Documentation → Integration → CSS API (beta)** renders it as one filterable table: by type, by status, and free-text search across names, values and replacements
- Consumers read it from the installed package, at the version installed — editor support, agents, lint rules. The Storybook page is the first consumer and the proof it is usable.

-----

### Custom properties — keyed object

`value` and `category` (`token` | `palette` | `commons` | `breakpoint`) are always present. `resolved` appears only when a `var()` chain resolves to something different.

```json
"--pr-t-font-heading-1-fontSize": {
  "value": "var(--pr-t-font-fontSize-350)",
  "category": "token",
  "resolved": "1.75rem"
},
"--commons-borderRadius-M": {
  "value": "4px",
  "category": "commons",
  "deprecated": true
}
```

### Utility classes — keyed object

`css` is an array of blocks. A class needs several when it targets more than one sub-selector or breakpoint.

```json
"pr-u-borderBottom0": {
  "css": [
    {
      "decls": "border-bottom: var(--pr-t-spacings-0) !important",
      "resolved": "border-bottom: 0 !important"
    }
  ],
  "deprecated": true,
  "replacement": "pr-u-borderBlockEnd0"
}
```

| Block field | What it means |
| --- | --- |
| `decls` | The CSS declarations. Always present. |
| `resolved` | `decls` with `var()` chains followed to concrete values, when that differs. |
| `sel` | The declarations apply to this pseudo-class or pseudo-element, not to the element itself — `::before`, `::after`, `:focus-visible`, `::-webkit-inner-spin-button`. |
| `media` | The `@media` condition the block sits under — a breakpoint (`(min-width: 20em)`, `not all and (min-width: 20em)`) or a feature query (`only print`, `(prefers-reduced-motion: no-preference)`). |
| `container` | Same, for `@container` — `(min-width: 20em)`, `not (min-width: 20em)`. |

**A block carrying `sel`, `media` or `container` applies only under that qualifier** — a class's unconditional effect is the blocks with none of the three. That distinction matters for anything answering "what does this class do".

### Mixins — an array, not a keyed object

All six fields are always present.

```json
{
  "name": "mask",
  "namespace": "a11y",
  "module": "commons/utils/a11y",
  "import": "@lucca-front/scss/src/commons/utils/a11y",
  "params": "$suffix: ''",
  "signature": "mask($suffix: '')"
}
```

19 of 56 take no parameters: `params` is then `""` and `signature` is just the name.

### Declared but not yet populated

| Category | Empty fields |
| --- | --- |
| Variables | `replacement`, `note` |
| Utilities | `note` |
| Mixins | `doc`, `deprecated`, `replacement`, `note` |

Utilities map cleanly — 394 of 484 deprecated ones carry a replacement. Variables mostly do not: `--colors-*` supersedes to the `--palettes-*` *family*, not to a name. Their replacements are left to the registry below rather than hand-authored.

## Deprecation data is temporary

- Comes from `css-api/deprecations.js`, a frozen regex seed for names generated inside SCSS loops that cannot carry an inline annotation
- Scoped to `packages/scss/css-api/` — **nothing here touches `packages/stylelint-config`**
- All lookups go through one function in `css-api/deprecation-source.js`, whose `kind` vocabulary mirrors the `deprecations.json` on `tech.scss.deprecation.strategy` (its `class` and `css-variable` names match manifest keys 1:1)
- Adopting that registry means rewriting one function body, and is a large upgrade: ~140 deprecated variables against the 32 detectable today
- No replacement data is authored by hand meanwhile — the registry supplies it properly

## Notes for reviewers

- **One deprecation contract across all three types** (`deprecated` / `replacement` / `note`). Replacement chains are collapsed at generation time — `u-textBrand` → `pr-u-textBrand` (also deprecated) → `pr-u-colorTextBrand`, so the manifest ships the final hop. Chains that dead-end have their replacement dropped instead: 90 deprecated classes ship without one. `selfCheck()` then guards the output invariant — all 394 surviving replacements name an existing, non-deprecated class — so it fires only if that resolution breaks.
- **The story reads a committed dataset, not `dist/`.** Storybook CI builds (`deploy`, `deploy-staging`, `e2e-test`) never run `npm run build`, and `build.js` wipes `dist/scss` — importing from `dist/` would fail CI. Same approach as `stories/documentation/icons-list.ts`. A CI step regenerates it and fails on any diff, so it cannot rot.
- **`build.js` never refreshes the committed dataset** — only `npm run css-api` does. A build must not dirty the working tree.
- The dataset is `linguist-generated`, so its 11k lines are collapsed in this diff.
- Expect adjacent-line conflicts with `tech.scss.deprecation.strategy` on `build.js`, `package.json` and `build.yml`; whichever lands second resolves them.

### Contribution

- [x] Designs are respected and all relevant options are handled — n/a, documentation page built from existing components
- [x] Responsive behavior addressed when needed
- [x] Feature is accessible (keyboard navigation, screen reader support, ARIA tags, etc.) — native table semantics with `scope="col"`, existing form components
- [x] Stories are updated and Storybook controls have descriptions
- [x] Feature is covered by E2E tests or UI diff — page carries the `id="ready"` sentinel
- [x] `npm run build` OK
- [x] `npm run lint` OK

### Functional review

- [ ] UI diff OK
- [ ] Feature and all options are manually tested and comply with design and guidelines.

-----


packages/scss/css-api/deprecations.js
```diff
@@ -0,0 +1,120 @@
+/**
+ * Frozen seed list of deprecated CSS API names.
+ *
+ * This exists ONLY for deprecated names that are generated by SCSS loops and
+ * therefore cannot carry an inline `/* @deprecated *\/` comment in the source
+ * (the primary, zero-maintenance mechanism — see generate.js `readDeprecatedComments`).
+ *
+ * Do not grow this list for new deprecations: mark those at the source with a
+ * loud `/* @deprecated *\/` comment on the declaration instead. This list is a
+ * one-time snapshot of the pre-existing deprecations that live only in loops.
```
MarcFairbrother: This is a temporary strategy for surfacing deprecated CSS APIs, we need to land on a single machine-readable source of truth for improved maintainability in the near future

stories/documentation/integration/css-api/css-api.stories.ts
```diff
@@ -0,0 +1,169 @@
...
+				white-space: nowrap;
+			}
+
+			.cssApi-kind {
+				white-space: nowrap;
+			}
+
+			.cssApi-value {
+				font-family: var(--pr-t-font-family-monospace, monospace);
+				font-size: 0.8125rem;
```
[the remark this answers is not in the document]
MarcFairbrother: 🤖 Produced by AI

Fixed in 6bb34e7da — now `font-size: var(--pr-t-font-body-S-fontSize)`. Note this is a small visual change, 0.8125rem to 0.875rem, since no token matched the hard-coded value.

stories/documentation/integration/css-api/css-api.stories.html
```diff
@@ -0,0 +1,107 @@
...
+				[ngModel]="search()"
+				(ngModelChange)="search.set($event ?? '')"
+				[ngModelOptions]="{ standalone: true }"
+				placeholder="Search…"
+				hasSearchIcon
+				hasClearer
+			/>
+		</lu-form-field>
+
+		<span *luFilterPillAddonAfter class="cssApi-results">{{ filtered().length }} result(s)</span>
```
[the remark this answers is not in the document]
MarcFairbrother: 🤖 Produced by AI

Fixed in 6bb34e7da.

stories/documentation/integration/css-api/css-api.stories.html
```diff
@@ -0,0 +1,107 @@
...
+	</lu-filter-bar>
+
+	@if (filtered().length === 0) {
+		<div class="cssApi-empty">No entry matches your search.</div>
+	} @else {
+		<div class="cssApi-scroll">
+			<table class="table mod-S">
```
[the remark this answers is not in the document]
MarcFairbrother: 🤖 Produced by AI

Fixed in 6bb34e7da. The page used `.table`, which has no sticky-header modifier — `mod-stickyHeader` lives on `indexTable` and `dataTable`. Since `docs/life-cycle.md` also lists `Table` as deprecated in favour of Data table / Index table, a new page shouldn't have been using it at all, so the table is now `.indexTable mod-stickyHeader` (a 1:1 class rename; `mod-S` only set the body-S font size, which the page's own styles cover).

Worth noting the header was one of two problems here: the table also rendered all 1755 rows, which is what timed out the Storybook test. It's now capped at 100 with a count of what's hidden, so the sticky header applies to a bounded list.

packages/scss/css-api/story-data.js
```diff
@@ -0,0 +1,125 @@
...
+		const blocks = entry.css || [];
+		// Conditional blocks are surfaced separately as responsive variants.
+		const base = blocks.filter((b) => !b.media && !b.container);
+		const conditional = blocks.filter((b) => b.media || b.container);
+		rows.push(
+			compact({
+				kind: 'utility',
+				name,
+				value: base.map(declText).join(' ') || blocks.map(declText).join(' '),
+				resolved: base.some((b) => b.resolved) ? base.map((b) => b.resolved || b.decls).join(' ') : undefined,
```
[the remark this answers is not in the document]
MarcFairbrother: 🤖 Produced by AI

Confirmed and fixed in 5eb3d7843. The `pr-u-onlyPrintDisplayBlock` case was worse than "unconditional": the value was `display: none`, i.e. the inverse of what the class does, because the base block won and the `@media only print` block carrying `display: block` was filtered out of the value entirely.

Every block is now rendered wrapped in its own qualifier:

```
pr-u-onlyPrintDisplayBlock  @media only print { display: block !important } display: none
pr-u-displayNoneAtMediaMinS @media (min-width: 50em) { display: none !important }
```

16 print utilities were inverted this way and 44 conditional-only ones read as unconditional. `variants` is kept as the scannable query summary.

On `pr-u-focusVisible`: the `:focus-visible` selector was already restored in `resolved` by 97b43c70b, which landed after the commit you reviewed — the old code dropped `sel` when building `resolved` but not when building `value`.

packages/scss/css-api/deprecations.js
```diff
@@ -0,0 +1,120 @@
+/**
+ * Frozen seed list of deprecated CSS API names.
+ *
+ * This exists ONLY for deprecated names that are generated by SCSS loops and
+ * therefore cannot carry an inline `/* @deprecated *\/` comment in the source
+ * (the primary, zero-maintenance mechanism — see generate.js `readDeprecatedComments`).
```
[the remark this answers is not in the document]
MarcFairbrother: 🤖 Produced by AI

Correct, and already fixed in 7dc6f17b6, which landed after the commit you reviewed. The helper is `readDeprecatedComment`. Its JSDoc had also been orphaned above the wrong function, and the "primary mechanism" framing was wrong in a second way: there are no loud `/* @deprecated */` comments anywhere in `packages/scss/src`, so that hook never fires and the seed list is the only source. The header now says so.

packages/scss/css-api/generate.js
```diff
@@ -0,0 +1,801 @@
...
+		}
+
+		const decls = collectDecls(rule);
+		if (decls === '') {
+			// still process (rare empty rule) but nothing to record
+		}
+
+		// Capture the shared animated attribute rule declarations.
+		for (const sel of rule.selectors) {
+			if (/\[class\*=['"](?:pr-)?u-animated['"]\]/.test(sel) && decls) {
```
F2X, line 406: Blocking — this capture never fires: dart-sass emits attribute-selector values unquoted, and the regex requires quotes around them.

Compiling that rule shape at this head gives:

```css
[class*=u-animated], [class*=pr-u-animated] {
  animation-duration: 0.3s;
  animation-fill-mode: both;
}
```

`/\[class\*=['"](?:pr-)?u-animated['"]\]/` does not match it, so `animatedShared` stays empty and the merge loop at 471-482 iterates nothing. The consequence is in the committed manifest: `pr-u-animatedFadeIn` ships as `animation-name: fadeIn`, with no duration, no timing function and no fill mode — so the manifest describes a class that would do nothing observable. All 48 `*-animated*` rows are affected, and no sentinel covers the path.

```suggestion
			if (/\[class\*=['"]?(?:pr-)?u-animated['"]?\]/.test(sel) && decls) {
```

packages/scss/css-api/generate.js
```diff
@@ -0,0 +1,801 @@
...
+	{ prefix: '--breakpoints-', category: 'breakpoint' },
+	{ prefix: '--commons-', category: 'commons' },
+	{ prefix: '--colors-', category: 'commons' }, // deprecated, still public surface
+	{ prefix: '--sizes-', category: 'commons' }, // deprecated, still public surface
+];
+
+const COMPONENT_VAR_RE = /^--components?-/;
+// The class-name group allows CSS-escaped characters (`\.`) so selectors like
+// `.pr-u-width100\%` are captured; the escape is stripped when the name is stored.
+const UTILITY_SELECTOR_RE = /^(?::root\s+)?\.((?:pr-)?u-(?:[A-Za-z0-9]|\\.)+)((?:::?)[A-Za-z-][\w-]*)?$/;
```
F2X, line 62: Blocking — the class-name group forbids a hyphen, so utilities carrying one are dropped with no warning.

`utils/index.scss:643` loops `core.$boxDirection`, which holds `'block-start'`, `'block-end'`, `'inline-start'`, `'inline-end'` (`core.scss:9`), and its `0` branch interpolates the direction raw — unlike the `spacing` mixin at `core.scss:44`, which passes it through `transform.camelize`. Compiling `packages/scss/src/main-all.scss` at this head yields 816 `u-*` selectors, 12 with an internal hyphen:

```
.pr-u-marginBlock-start0   .pr-u-paddingBlock-start0   .pr-u-insetBlock-start0
.pr-u-marginBlock-end0     .pr-u-paddingBlock-end0     .pr-u-insetBlock-end0
.pr-u-marginInline-start0  .pr-u-paddingInline-start0  .pr-u-insetInline-start0
.pr-u-marginInline-end0    .pr-u-paddingInline-end0    .pr-u-insetInline-end0
```

`grep -c 'marginBlock-start' stories/documentation/css-api-manifest.ts` returns 0 — none of the 1 755 rows carry them, and with the `u-*` twins the manifest is generated with that is 24 missing rows. `selfCheck` cannot catch it: it asserts `>=750 pr-u-*`, a floor the output clears either way. So the guarantee that the artifact "cannot drift from what the library ships" does not hold on merge.

Allowing the hyphen still matches every other shape — I checked the escaped (`.pr-u-width100\%`), pseudo-class and pseudo-element forms against the fixed pattern.

```suggestion
const UTILITY_SELECTOR_RE = /^(?::root\s+)?\.((?:pr-)?u-(?:[A-Za-z0-9-]|\\.)+)((?:::?)[A-Za-z-][\w-]*)?$/;
```

packages/scss/css-api/story-data.js
```diff
@@ -0,0 +1,151 @@
...
+ * @param {(block: object) => string} pick which field to read from each block
+ */
+function joinBlocks(blocks, pick = (block) => block.decls) {
+	return blocks.reduce((out, block) => {
+		const text = blockText(block, pick);
+		if (!out) {
+			return text;
+		}
+		const braced = Boolean(block.sel || block.media || block.container);
+		return out + (braced || out.endsWith('}') ? ' ' : '; ') + text;
```
F2X, line 100: Blocking — the separator is read off the incoming block, so a bare declaration list followed by a braced one loses its terminating semicolon.

The JSDoc above states the rule, and it is the rule that is wrong: a braced group is "already delimited" on its right, never on its left. `braced` inspects the block being appended and `out.endsWith('}')` catches only a previous braced block, so bare-then-braced falls through to `' '`. In the committed manifest:

```css
pr-u-summaryReset: list-style: none !important ::-webkit-details-marker { display: none !important }
```

That does not parse. The Value column is the part of this page a consumer copies, so the defect lands exactly where the page is meant to be trusted.

```suggestion
		return out + (out.endsWith('}') ? ' ' : '; ') + text;
```

The JSDoc above needs the same correction — a braced group is self-delimiting on its right only.

packages/scss/css-api/story-data.js
```diff
@@ -0,0 +1,151 @@
...
+				import: mixin.import,
+				doc: mixin.doc,
+				deprecated: mixin.deprecated,
+				replacement: mixin.replacement,
+				note: mixin.note,
+			}),
+		);
+	}
+
+	rows.sort((a, b) => a.kind.localeCompare(b.kind) || a.name.localeCompare(b.name));
```
F2X, line 82: Blocking — `localeCompare` with no locale makes the row order depend on the generating machine's ICU data, and `build.yml:30` byte-compares the result.

The committed file is in collated order, which is not code-unit order: `pr-u-clearfix` (line 984) precedes `pr-u-clearInlineEnd` (line 989), while comparing code units puts `clearInlineEnd` first (`'I'` 0x49 < `'f'` 0x66). The artifact therefore encodes the collation of whoever last ran the generator.

A contributor on a Node built without full ICU, or on a Node whose CLDR revision differs from the runner's, regenerates a file that differs only in row order — and the gate fails with "css-api-manifest.ts is stale" on semantically identical content, for a file they are told not to hand-edit. The fix is a comparator that does not read the ambient locale.

```suggestion
	rows.sort((a, b) => (a.kind < b.kind ? -1 : a.kind > b.kind ? 1 : a.name < b.name ? -1 : a.name > b.name ? 1 : 0));
```

packages/scss/css-api/generate.js
```diff
@@ -0,0 +1,801 @@
...
+	const categories = new Set(Object.values(manifest.variables).map((v) => v.category));
+	for (const expected of ['token', 'palette', 'breakpoint', 'commons']) {
+		if (!categories.has(expected)) {
+			errors.push(`No variables in category "${expected}"`);
+		}
+	}
+
+	// Value sentinels.
+	const brand500 = manifest.variables['--palettes-brand-500'];
+	if (!brand500 || (brand500.resolved || brand500.value) !== '#ff9361') {
```
F2X, line 683: This sentinel hard-codes a design value, and `generateCssApi()` is now on the critical path of four pipelines.

`build.js:45` calls it inside `npm run build`, and this PR adds `npm run css-api` to `deploy.yml:78`, `deploy-staging.yml:55` and `e2e-test.yml:20`. So the day the brand ramp moves, `--palettes-brand-500 sentinel failed` turns the package build, both deploys and the e2e job red together, and the remedy is editing a hex literal in a documentation generator.

The count floors fail in the opposite direction: `>=200 deprecated u-* classes` (line 656) and `>=750 pr-u-*` (line 653) break the build when the deprecation campaign this manifest exists to serve actually deletes classes.

Keep the structural sentinels — the component-var leak check, category coverage, the escaped selector, the media block, the derived replacement. Drop the ones asserting a value the design system owns, or a count meant to shrink. A generator that documents the surface should not be able to veto a change to it.

packages/scss/css-api/deprecations.js
```diff
@@ -0,0 +1,125 @@
...
+	// Deprecated font-size utility classes generated by the deprecated `sizes` mixin.
+	{
+		test: /^pr-u-text(XS|S|M|L|XL|XXL|XXXL)$/,
+	},
+	// Deprecated border-radius utilities generated by the deprecated `borderRadius`
+	// mixin. Corner variants exist for M / L / XL / Full; the corner-less form is
+	// deprecated only for M / L / XL — `pr-u-borderRadiusFull` (and other tokens
+	// like Structure/Default) are the modern token classes and stay non-deprecated.
+	{
+		test: /^pr-u-border(?:[A-Z][a-z]+[A-Z][a-z]+Radius(?:M|L|XL|Full)|Radius(?:M|L|XL))$/,
```
F2X, line 99: The 32 corner variants this rule deprecates have no migration target, and none exists in the surface.

The comment above holds for the corner-less form: `pr-u-borderRadiusFull` declares both tokens with the modern `--pr-t-border-radius-full` last, so it is a migrated class and staying non-deprecated is right. The gap is on the other side of the rule.

It matches 35 names, 32 of them corner variants (`pr-u-borderBottomLeftRadiusFull`, `…RadiusM`, `…RadiusL`, `…RadiusXL`, across 8 corners). None carries a `replacement`, and the successor shape does not exist either: `pr-u-borderRadiusStructure` is in the manifest, `pr-u-borderStartStartRadiusStructure` and its 31 siblings are not.

A lint rule built on this manifest would emit 32 warnings a reader cannot act on. Either point them explicitly at the corner-less token classes, or leave them out of the deprecated set until the corner successors land.

stories/documentation/integration/css-api/css-api.stories.html
```diff
@@ -0,0 +1,121 @@
...
+								}
+							</td>
+							<td class="indexTable-body-row-cell">
+								<span class="cssApi-kind">{{ label(entry.kind) }}</span>
+								@if (entry.group) {
+									<span class="cssApi-meta">{{ entry.group }}</span>
+								}
+							</td>
+							<td class="indexTable-body-row-cell">
+								<span class="cssApi-value">{{ entry.resolved || entry.value }}</span>
```
F2X, line 97: The cell renders `resolved || value` while the filter matches `value`, so a token search returns rows where the searched string appears nowhere on screen.

`css-api.stories.ts:155` tests `entry.value`. Search `--pr-t-spacings-100`: `pr-u-marginBlockStart100` matches, and its cell shows the resolved form `margin-block-start: 0.5rem !important`. Which token a utility reads is the thing a consumer most needs from a CSS API — it is what makes a token change predictable — and once `resolved` exists it is the one field the page never shows.

Render the authored value and demote `resolved` to the secondary line, the way `cssApi-meta` already handles `group` and `variants`.

```suggestion
								<span class="cssApi-value">{{ entry.value }}</span>
								@if (entry.resolved) {
									<span class="cssApi-meta">{{ entry.resolved }}</span>
								}
```

packages/scss/css-api/generate.js
```diff
@@ -0,0 +1,801 @@
...
+	/** @type {Map<string, { value: string, category: string, deprecated?: boolean, note?: string }>} */
+	const variables = new Map();
+
+	root.walkDecls((decl) => {
+		if (!decl.prop.startsWith('--')) {
+			return;
+		}
+		// Record every custom property (even component/internal ones) so var()
+		// chains referencing them can resolve.
+		if (!rawValues.has(decl.prop)) {
```
F2X, line 339: First-seen-wins makes every `resolved` field in the manifest a function of source order in `main-all.scss`.

`rawValues` records the first textual occurrence of a custom property, and `root.walkDecls` reaches component-scoped and `@media`-nested declarations too. Compiling the surface at this head, `--palettes-brand-500` is declared three times:

```css
/* 3352 */ :root { --palettes-brand-500: #ff9361; }
/* 3670 */ :root { --palettes-brand-500: var(--palettes-brandContrasted-500); }
/* 3732 */ :root { --palettes-brand-500: var(--palettes-brandContrasted-500); }
```

It resolves correctly today only because the canonical declaration happens to be first. Move `@layer utils` above the palette block, let a component declare a `--pr-t-*` locally, or reorder the entry points, and every `resolved` value silently switches to the override chain — with no sentinel that would notice, since the only resolved sentinel (line 707) just checks the string holds no `var(`.

Resolve against the `variables` map — the canonical, non-conditional `:root` values you already collect — rather than first-seen raw text.

F2X (changes requested):
### Reviewed changes

Reviewed 14 of 15 changed files and left 8 comments.

<details>
<summary>Show a summary per file</summary>

| File | Description |
| --- | --- |
| `.gitattributes` | Marks the generated manifest so it is collapsed in diffs. |
| `.github/workflows/build.yml` | Adds a staleness gate comparing the committed manifest to a fresh generation. |
| `.github/workflows/deploy-staging.yml` | Regenerates the CSS API before the staging Storybook build. |
| `.github/workflows/deploy.yml` | Regenerates the CSS API before the production Storybook build. |
| `.github/workflows/e2e-test.yml` | Regenerates the CSS API before the e2e job. |
| `build.js` | Runs the manifest generation as a step of the package build. |
| `package.json` | Adds the `css-api` script that drives the generator. |
| `packages/scss/css-api/deprecation-source.js` | Reads deprecation markers out of the SCSS sources. |
| `packages/scss/css-api/deprecations.js` | Holds the seed rules naming deprecated utilities and custom properties. |
| `packages/scss/css-api/generate.js` | Compiles the SCSS surface with postcss and emits the manifest. |
| `packages/scss/css-api/story-data.js` | Flattens the manifest into the rows the story renders. |
| `stories/documentation/css-api-types.ts` | Declares the row type the story consumes. |
| `stories/documentation/integration/css-api/css-api.stories.html` | Renders the filterable CSS API table. |
| `stories/documentation/integration/css-api/css-api.stories.ts` | Holds the story component, its filters and its styles. |

</details>

<details>
<summary>Files not reviewed (1)</summary>

* **`stories/documentation/css-api-manifest.ts`**: Generated dataset, 11 331 lines — parsed as data to check the findings below, not read line by line.

</details>

### Not anchored to a line

| Status | Point | Ask |
| --- | --- | --- |
| ⚠️ **Warning** | deploy, deploy-staging and e2e now all run `npm run css-api`, so no CI consumer reads the committed 11 331-line manifest any more — only local `npm start` does. | Consider a `prestart` script instead of the committed copy plus its gate. |
| ❌ **Missing** | 1 151 added JS lines are in neither lint nor tests: `lint:es` is `eslint --quiet packages/` and eslint.config.mjs declares no `**/*.js` block. | Add a `files: ['**/*.js']` block, so `selfCheck` is not the only guard. |
| ℹ️ **Info** | `stories/qa/tokens-explorer` has the same filter shell and documents the same 617 custom properties from a hand-maintained list. | Point it at this dataset, or retire it. |
| ✅ **Checked** | No `resolved` field leaks an unresolved `var()` — 0 of 1 755 rows, a failure mode I suspected and refuted, though the sentinel covers only one row. | No change needed. |

<sub>— via Claude Code</sub>

PR #5237, opened 2026-08-11
fix(filter-pills): stop the select loader from sticking on the second pill opening

### LINKED ISSUE #5236: [FilterPill] API select shows a stuck "Loading..." row from the second opening
### Bug description

Since 21.4.1, an API-fed select (`lu-multi-select [departments]`, and any other `ALuCoreSelectApiDirective`-based select) hosted in a **filter pill** shows a permanent "Loading..." row at the bottom of its panel from the **second opening** onwards, even when there is nothing left to load. The options themselves display fine — only the loader never turns off.

### How to reproduce

1. Put an API select in a filter bar, e.g. `<lu-multi-select departments />` inside a `<lu-filter-pill>`.
2. Open the pill, wait for the options to load — everything is fine.
3. Close the popover by clicking outside.
4. Open the pill again and scroll to the bottom of the panel: a "Chargement en cours..." row is displayed forever.

### Root cause

Two things combine:

- `FilterPillComponent` only binds `(luPopoverOpened)`, not `(luPopoverClosed)`. When the popover closes on its own (outside click — `luPopoverNoCloseButton` is set, so that is the nominal way to close it), `onFilterPillClosed()` is never called and the select's `isPanelOpen$` stays `true`.
- Since #5172 (`3ad89bf01`, shipped in v21.4.1), `ALuCoreSelectApiDirective.buildOptions()` turns the loader on with a `tap` on **every** `isPanelOpen$` emission, while the fetch that turns it back off only runs on a closed → open transition (`pairwise()` + `distinctUntilChanged()` downstream). On the second opening the pill re-emits `true`: the loader lights up, no fetch runs, nothing ever resets `loading`.

Before #5172 the desync was invisible; now it leaves the loader on forever for every API select in a filter pill.

### Expected behavior

Closing the pill popover by any means should notify the select (`isPanelOpen$` back to `false`), and reopening it should behave like the first opening — loader on while fetching, off when done.

Repro'd in Talent.PeopleReview (LF 21.4.1, department & establishment multi-selects of a filter bar). Fix PR incoming.

lmullot:
## Description

An API select hosted in a filter pill no longer shows a stuck "Loading..." row at the bottom of its panel from the second opening onwards.

Closes #5236

-----

The pill template only bound `(luPopoverOpened)`: a popover closed from its own side (outside click, Escape) never called `onFilterPillClosed()`, so the select's `isPanelOpen$` stayed `true`. Since #5172, `ALuCoreSelectApiDirective` turns the loader on upon every `isPanelOpen$` emission while the fetch that turns it off only runs on a closed to open transition. On the second opening the pill re-emitted `true`: the loader lit up and nothing ever reset it.

Two complementary changes:

- `FilterPillComponent` now binds `(luPopoverClosed)` to notify the input, which covers every close path. `closePopover` no longer calls `onFilterPillClosed()` itself: the popover event is the single source of truth.
- `ALuCoreSelectApiDirective` ignores re-emitted open states (`distinctUntilChanged()` before the loader tap), so a desynced panel state can no longer strand the loader.

How to test: put `<lu-multi-select departments />` in a `<lu-filter-pill>`, open the pill, close it by clicking outside, open it again and scroll to the bottom of the panel. Before the fix a permanent "Loading..." row shows; after it the panel behaves like the first opening.

-----

### DIFF
--- a/packages/ng/core-select/api/api.directive.spec.ts
+++ b/packages/ng/core-select/api/api.directive.spec.ts
@@ -116,6 +116,26 @@ describe('ALuCoreSelectApiDirective', () => {
 		expect(select.loading$.value).toBe(false);
 	}));
 
+	it('should not restart the loader when the open state is re-emitted without a close', fakeAsync(() => {
+		fixture.detectChanges();
+		tick(); // Component initialization uses a setTimeout :see_no_evil:
+
+		select.openPanel();
+		fixture.detectChanges();
+		tick(); // openPanel defers its work in a setTimeout
+		tick(MAGIC_OPTION_SCROLL_DELAY);
+
+		expect(select.loading$.value).toBe(false);
+
+		// A filter pill whose popover closed without notifying the select re-emits `true` on the
+		// next opening: no fetch runs on an open → open transition, so nothing would reset the loader
+		select.onFilterPillOpened();
+		fixture.detectChanges();
+		tick(MAGIC_OPTION_SCROLL_DELAY);
+
+		expect(select.loading$.value).toBe(false);
+	}));
+
 	it('should query options once when searching while the select is closed', fakeAsync(() => {
 		tick(); // Component initialization uses a setTimeout :see_no_evil:
 		select.clueChanged('test');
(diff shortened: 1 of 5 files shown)


lmullot:
rebuild

Supamiu (approved):

PR #5241, opened 2026-08-14
[tooltips] fix safari behaviours

### LINKED ISSUE #5240: [lu-tooltip] always displayed on Safari even with luTooltipWhenEllipsis
### Bug description

On Safari, when i use the tooltip with the `luTooltipWhenEllipsis` directive, the tooltip is always displayed on mouse hover even without ellipsis, and the element is focusable.

https://github.com/user-attachments/assets/1b99ffef-4bb0-4123-acc3-0e4d4423563e

### How to reproduce
Use this code on Safari and just mouse hover or navigate into with keyboard.
`<span class="pr-u-ellipsis" luTooltipWhenEllipsis luTooltip>Lucca Front</span>`

vvalentin-lucca:
## Description

Two issues resolved: 
- The `[luTooltipWhenEllipsis]` attribute selector was not robust enough, as the attribute was sometimes removed from the DOM. It has now been replaced by a class.
- In Safari, the width calculation was incorrect for `inline` elements, causing behavior that differed from other browsers. 

-----

Fix #5240 

-----

### DIFF
--- a/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
+++ b/packages/ng/tooltip/trigger/tooltip-trigger.directive.ts
@@ -187,7 +187,7 @@ export class LuTooltipTriggerDirective implements OnDestroy {
 				}
 				const host = this.#host.nativeElement;
 				const hostStyle = getComputedStyle(host);
-				// No need to run a test if the element isn't an ellipse
+				// No need to run a test if the element is not truncated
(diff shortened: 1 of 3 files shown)


martincadoux (approved):

ymalie-lucca (approved):

Supamiu (approved):

Supamiu:
Tests won't pass for some reason oO

PR #5246, opened 2026-08-19
fix(table-skeleton): colsAlign index start at 1

ymalie-lucca:
## Description

🔥  **Breaking**
The `colsAlign` keys on `lu-skeleton-table`, `lu-skeleton-index-table` and `lu-skeleton-data-table` are now column numbers starting at `1` instead of `0`-based indexes.
Increment every key by one: `{ 0: 'center', 1: 'end' }` becomes `{ 1: 'center', 2: 'end' }`.

-----


Supamiu:
Can you give us a bit more details on why this is needed please? We use 0-based indexes everywhere, using 1 just here feels wrong.

ymalie-lucca:
@Supamiu I already discussed this with Jérémie a couple of months ago, it's so uncomfortable when I know that the fourth column of the table needs to be aligned to the right. I have to think about how it's calculated in the component code, then subtract one to set the correct value that matches the code logic rather than the human reading logic.

jeremie-lucca (approved):

Issue #5248, opened 2026-08-19
[Resource card] cards overflow from their wrapper in grid mode


martincadoux:
### Bug description
In grid mode, resource cards overflow from their wrapper when it is below 336 pixels wide. This creates a non-conformity to RGAA's 10.11 criteria.

This is due to the fixed minimum width of the card set to 20rem.

<img width="321" height="569" alt="Image" src="https://github.com/user-attachments/assets/87351266-32eb-4a6d-a0e5-9fa7db7b3f2e" />

### How to reproduce
- Set a `<lu-resource-card-wrapper>` with the `grid` input,
- add on or multiple `<lu-resource-card>` inside,
- set the viewport width at 320px.

PR #5250, opened 2026-08-20
[ressourceCard] fix wrapper min inline size

vvalentin-lucca:
## Description

The minimum width of the wrapper in grid mode used to depend solely on the minimum size of the cards. Now it also depends on the width of the parent container, and the smaller of the two takes precedence.

-----

#5248 

Thanks to @martincadoux  for the fix.

-----


martincadoux (approved):
Merci !👌

Supamiu (approved):

PR #5261, opened 2026-08-21
[tag] hiddenLabel or onlyIcon

### LINKED ISSUE #5217: [Tag] Allow icon only
### Intention
Allow a Tag variation allowing only an icon

### Design
- Height / spacing should stay the same 

### Key points
It's currently technically possible to display and icon without label, but the UI is broken:

Image

vvalentin-lucca:
## Description

New option for tags: use them only with an icon.
Either without associated text (purely decorative), or with associated text that is hidden (informational).

-----

The component will also detect if it contains neither text nor an icon, in which case it will no longer be displayed.

Close #5217 

-----
Capture d’écran 2026-08-31 à 16 24 10

### DIFF
--- a/packages/scss/src/components/tag/index.scss
+++ b/packages/scss/src/components/tag/index.scss
@@ -41,5 +41,15 @@
 		&:has(.lucca-icon) {
 			@include withIcon;
 		}
+
+		&:has(.tag-content:empty) {
+			&:not(:has(.lucca-icon)) {
+				@include empty;
+			}
+		}
+
+		&:empty {
+			@include empty;
+		}
 	}
 }
--- a/stories/qa/tags/tags.stories.html
+++ b/stories/qa/tags/tags.stories.html
@@ -63,7 +63,6 @@
 						<span class="tag mod-AI">
 							<span class="lucca-icon icon-heart" aria-hidden="true"></span> <span class="tag-content">Text</span>
 						</span>
-
 						<span class="tag mod-L palette-product">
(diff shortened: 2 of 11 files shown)


packages/scss/src/components/tag/index.scss
```diff
@@ -41,5 +41,15 @@
...
+
+		&:has(.tag-content:empty) {
+			&:not(:has(.lucca-icon)) {
+				@include empty;
+			}
+		}
+
+		&:empty {
+			@include empty;
+		}
```
fbasmaison-lucca, line 53: Isn’t it a state more than a mod?

stories/qa/tags/tags.stories.html
```diff
@@ -92,6 +91,38 @@
...
+					<span class="lucca-icon icon-heart" aria-hidden="true"></span>
+					<span class="tag-content pr-u-mask">Text</span>
+				</span>
+			</td>
+			<td>
+				<lu-tag icon="heart" label="Text" hiddenLabel />
+			</td>
+		</tr>
+		<tr>
+			<td>Empty</td>
```
fbasmaison-lucca, line 118: Do we have a use for this case? :)

fbasmaison-lucca (dismissed):
Just two questions, including one out of curiosity.

stories/qa/tags/tags.stories.html
vvalentin-lucca: No, it's more to ensure that the display works properly in case the component is configured incorrectly.

jeremie-lucca (dismissed):

jeremie-lucca (approved):

PR #5264, opened 2026-08-21
feat(form): focus the first invalid field on submit

guillerot:
## Description

`form[luForm]` can now move focus to the first invalid field when the form is submitted while invalid, so keyboard and screen reader users are led straight to the error (WCAG 3.3.1). The behavior is opt-in via a new `focusInvalidOnSubmit` boolean input (default `false`).

-----

The behavior is DOM-based on purpose: after the `submit` event, `afterNextRender` lets the app's own submit handler run first (typically `markAllAsTouched`) and change detection apply the validity state; then, if the host `<form>` carries `ng-invalid`, the innermost `.ng-invalid` element is resolved and focused. When that element is not natively focusable (custom control hosts like `lu-simple-select`), the first focusable descendant is focused instead, then scrolled into view.

Working with CSS state classes rather than injecting `FormGroupDirective`/`NgForm` keeps it agnostic: it covers reactive forms and `ngModel` alike, and needs no wiring in consuming apps.

Points to discuss before undrafting:

- No story yet: I can add a Storybook story demonstrating the behavior if you confirm the approach.

Validated end-to-end in Cleemy Procurement with the same logic as a local directive (supplier creation form: submit via click and via Enter both land focus on the first invalid field).

-----

### Contribution

- [ ] Designs are respected and all relevant options are handled
- [ ] Responsive behavior addressed when needed
- [x] Feature is accessible (keyboard navigation, screen reader support, ARIA tags, etc.)
- [ ] Stories are updated and Storybook controls have descriptions
- [ ] Feature is covered by E2E tests or UI diff
- [x] `npm run build` OK
- [x] `npm run lint` OK

### Functional review

- [ ] UI diff OK
- [ ] Feature and all options are manually tested and comply with design and guidelines.

-----

### DIFF
--- a/packages/ng/form/form.component.ts
+++ b/packages/ng/form/form.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, ElementRef, forwardRef, inject, input, ViewEncapsulation } from '@angular/core';
+import { afterNextRender, ChangeDetectionStrategy, Component, ElementRef, forwardRef, inject, Injector, input, ViewEncapsulation } from '@angular/core';
 import { luBooleanAttribute } from '@lucca-front/ng/core';
 import { LU_FORM_INSTANCE } from './form-instance';
 import { LuDialogRef } from '@lucca-front/ng/dialog';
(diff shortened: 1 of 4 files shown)


packages/ng/form/form.component.ts
```diff
@@ -26,7 +28,23 @@ import { LuDialogRef } from '@lucca-front/ng/dialog';
...
 
 	readonly presentation = input(false, { transform: luBooleanAttribute });
+
+	readonly focusInvalidOnSubmit = input(true, { transform: luBooleanAttribute });
+
+	protected onSubmit(): void {
+		if (!this.focusInvalidOnSubmit()) {
+			return;
+		}
+		setTimeout(() => {
```
JeSuisCharlie1: can we use `afterNextRender` instead ?

JeSuisCharlie1 (dismissed):

Supamiu:
Pushing it our of 22.0 so we can properly discuss it and find the best approach.

JeSuisCharlie1 (approved):

jeremie-lucca (approved):

PR #5286, opened 2026-08-28
fix(dialog): keep a routed dialog open on history navigation while its route survives

batnyu:
## Description

A routed dialog no longer closes on a browser back/forward when its route stays active — e.g. the parent dialog of a nested dialog route stayed on screen with no dialog after a history back.

https://github.com/user-attachments/assets/4a003ad5-2042-4516-a059-3ac14efde6a1

-----

Routed dialogs now open with `closeOnNavigation: false`: their lifecycle already belongs to the router (`DialogRoutingContainerComponent` dismisses them when their route deactivates), while the cdk's default disposed every overlay on any popstate. When a history navigation left the dialog's route active (back from a nested dialog route, or a query-params-only back), the overlay was destroyed with no route deactivation to recreate it: depending on timing the dialog either silently closed and reopened (losing its state), or stayed gone with its route still active (with `withViewTransitions`, the recovery navigation loses the race and gets skipped).

Side effect worth noting: on a back that *does* deactivate the dialog's route, `onDismissed` now runs with the `DIALOG_ROUTE_DISMISS_TRIGGER` set to `navigation` (the container drives the dismissal) instead of `dialog:dismissed` (the cdk disposal used to win the race and masquerade as a user dismissal).

A caller passing `cdkConfigOverride: { closeOnNavigation: true }` still takes precedence.

New "Routing nested" stories cover the scenario (a detail dialog with a nested sub-dialog route, closed through a history back), with and without the router's view transitions. They use `withHashLocation` so the back stays a same-document popstate inside the storybook iframe.

-----

### Contribution

- [x] `npm run build` is OK
- [x] `npm run lint` is OK

### Functional review

- [ ] UI diff is OK

-----

### DIFF
--- a/packages/ng/dialog/dialog-routing/dialog-routing.component.ts
+++ b/packages/ng/dialog/dialog-routing/dialog-routing.component.ts
@@ -124,6 +124,8 @@ export class DialogRoutingContainerComponent<C> implements OnDestroy, OnInit {
 
 		this.#ref = this.#dialog.open<C>({
 			...dialogConfig,
+			// A routed dialog closes through its route, not through the cdk's closeOnNavigation.
+			cdkConfigOverride: { closeOnNavigation: false, ...dialogConfig.cdkConfigOverride },
 			content: this.dialogTemplate(),
 			data,
 			canClose: this.#getCanCloseFn(dialogConfig),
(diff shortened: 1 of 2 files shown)


JeSuisCharlie1 (dismissed):

jeremie-lucca:
@batnyu I got a weird behavior. The history back button seems to open a new dialog, overlapping the first one.

https://github.com/user-attachments/assets/58c24398-e998-456f-b6bc-cc0965700038

batnyu:
@jeremie-lucca the bug was caused by the fact that we instantiate 2 routing in the same page. I fixed it here https://github.com/LuccaSA/lucca-front/pull/5286/changes/8c2f46e11cb2e9893f31e763cc1ff43b02175783 by putting each story in an iframe. Let me know if this fix is acceptable

JeSuisCharlie1 (dismissed):

jeremie-lucca (dismissed):

jeremie-lucca (approved):

PR #5311, opened 2026-09-07
fix(core-select): detect user homonyms across pages

GuillaumeNury:
## Description

In user selects, two homonyms only got their department as additional information when they showed up on the same page of results. The last user of a page and the first of the next one therefore stayed indistinguishable. Detection now runs across all loaded pages.

-----

Homonyms resolution ran in `getOptionsPage`, hence one page at a time, while pages are accumulated downstream in `buildOptionsFromDataSource`.

- `SelectDataSource` exposes an optional `mapLoadedOptions(options)` hook, applied by `buildOptionsFromDataSource` on the whole list of loaded pages;
- `ALuCoreSelectApiDirective` wires it to an overridable `mapLoadedOptions` method (identity by default);
- `LuCoreSelectUsersDirective` implements it with `handleHomonyms`, removed from `getOptionsPage`.

Second commit, on the service cache: users without department were cached as an empty string, which the `!this.cache[id]` lookup read as a miss — their department was refetched on every resolution (twice per page now that resolution runs on the whole list). Ids missing from the response are now cached too, and concurrent resolutions share a single in-flight request. Net result: no extra HTTP call compared to before.

-----

### Contribution

- [x] `npm run build` is OK
- [x] `npm run lint` is OK

### Functional review

- [ ] UI diff is OK

-----

### DIFF
--- a/packages/ng/core-select/api/api.directive.ts
+++ b/packages/ng/core-select/api/api.directive.ts
@@ -75,6 +75,7 @@ export abstract class ALuCoreSelectApiDirective<TOption, TParams = Record<string
 			clueDebounceMs: this.debounceDuration,
 			getTotalCount: () => this.totalCount$,
 			reset: () => this.clearLastPageByClue(),
+			mapLoadedOptions: (options) => this.mapLoadedOptions(options),
 			getOptions: ({ clue, page }) => {
 				const lastPage = clue === this.#lastClue ? this.#lastPage : undefined;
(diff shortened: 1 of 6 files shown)


JeSuisCharlie1:
This is a duplicate of this PR : https://github.com/LuccaSA/lucca-front/pull/5303 ?

GuillaumeNury:
@JeSuisCharlie1 totally!

PR #5313, opened 2026-09-07
feat(select): open panel as dialog on smaller screens

Supamiu:
## Description

Select now opens in a dialog when display it too small.

Partially made with Claude, especially for the base structure, as it's basically "reuse what I did for filter pill but on a dialog".

I opened a draft so we can check for conflicts and have a public storybook to play with

-----

-----


JeSuisCharlie1 (dismissed):

packages/ng/core-select/panel/panel-sheet.ts
```diff
@@ -0,0 +1,95 @@
...
+function trackVisibleViewport<TPanel>(dialogRef: LuDialogRef<TPanel, never>): void {
+	const viewport = window.visualViewport;
+	if (!viewport) {
+		return;
+	}
+
+	const update = (): void => {
+		const style = document.documentElement.style;
+		style.setProperty('--components-dialog-visibleViewportHeight', `${viewport.height}px`);
+		style.setProperty('--components-dialog-visibleViewportBottomOffset', `${window.innerHeight - viewport.height - viewport.offsetTop}px`);
```
Supamiu: Wouldn't this affect existing dialogs too? Say we have a select inside a dialog, because it's a generic var, I assume it's used by every dialog.

vvalentin-lucca: I'm pushing one more commit.

PR #5316, opened 2026-09-08
feat(date-input): add humanized input to display dates in natural language

JeSuisCharlie1:
## Description

`lu-date-input` accepts a new `humanized` input. When enabled, the selected date is displayed in natural language while the field is not focused: "Yesterday", "Today" or "Tomorrow" in `day` mode, and the month name with the year ("March 2026") in `month` mode. Any other date, as well as `week` and `year` modes, keeps its numeric format.

Wording comes from `Intl`, so it is localized in every locale without any translation key. The option is opt-in and defaults to `false`, so existing fields are unchanged. `lu-date-range-input` is not covered.

-----

The tests are done with Claude 🤖 

-----


JeSuisCharlie1:
- Ne pas faire le switch de display quand le panel est ouvert ✅

jeremie-lucca (approved):

PR #5331, opened 2026-09-10
fix(rich-text-input): fix tags copy/paste and undo in plain-text mode

guillerot:
## Description

Pasting a copied tag in a plain-text rich text input (an email subject, for instance) recreates the tag chip. It used to paste `{{Tag}}` as raw text. Undoing the deletion of a tag no longer breaks the editor, which used to throw `Reconciliation: could not find DOM element` on every following keystroke.

-----

Reported in LuccaSA/Talent.PeopleReview#1645.

Start with `plain-text-formatter.ts`. In plain-text mode Lexical inserts pasted content as raw text, and the formatter only applied its `{{Tag}}` transformers when parsing the initial value. A `TextNode` transform now applies them on every text change, so typing `{{tag}}` by hand also creates a chip.

Then `tag-node.ts`. The chip `ComponentRef` was stored on the node instance and destroyed by `remove()`. Lexical reuses old node instances on undo, so restoring a deleted tag reused a destroyed component and threw NG0953 during the reconciliation. Chips are now created on every `createDOM`, tracked per editor and node key, and the stale ones are destroyed once the DOM is reconciled. This also fixes the chip label being duplicated when its DOM was recreated. The two new specs cover the paste and the undo scenarios.

Stories to check in the Storybook preview: `RichTextInput / Angular / With Tag Plugin Plain Text` and `With Tag Plugin Markdown` (copy, paste, Ctrl+A, delete, undo, redo).

-----

### Contribution

- [x] `npm run build` is OK
- [x] `npm run lint` is OK

### Functional review

- [ ] UI diff is OK

-----

### DIFF
--- a/packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
+++ b/packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
@@ -41,7 +41,7 @@ function getTagChips(editor: LexicalEditor): Map<NodeKey, Set<ComponentRef<ChipC
 }
 
 /** Destroy the chip components of a tag node that are no longer in the editor DOM. */
-export function destroyStaleTagChips(editor: LexicalEditor, nodeKey: NodeKey): void {
+function destroyStaleTagChips(editor: LexicalEditor, nodeKey: NodeKey): void {
 	const chips = getTagChips(editor);
 	const refs = chips.get(nodeKey);
 	if (!refs) {
--- a/packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
+++ b/packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
@@ -1,5 +1,6 @@
 import {
 	$createTextNode,
+	$getNodeByKey,
 	COMMAND_PRIORITY_NORMAL,
 	DecoratorNode,
 	type DOMConversion,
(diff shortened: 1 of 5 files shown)


packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
```diff
@@ -27,12 +28,48 @@ export type SerializedTagNode = Spread<
...
+	let chips = tagChips.get(editor);
+	if (!chips) {
+		chips = new Map();
+		tagChips.set(editor, chips);
+	}
+	return chips;
+}
+
+/** Destroy the chip components of a tag node that are no longer in the editor DOM. */
+export function destroyStaleTagChips(editor: LexicalEditor, nodeKey: NodeKey): void {
```
obraud: Pas forcément besoin d'exporter cette fonction qui semble interne?

obraud (dismissed):

packages/ng/forms/rich-text-input/plugins/tag/tag-node.ts
guillerot: 👍

obraud (approved):

jeremie-lucca (approved):

